Issue #5, opened 2024-09-24
Site selector state management


BaileyChessum:
The site selector component is used in a variety of places for science GUI. 

![Image](https://github.com/user-attachments/assets/5fe1075e-e82f-4553-8fd0-8daa6712208f)

![Image](https://github.com/user-attachments/assets/47c76204-94cb-4fed-b17e-abc8637e270f)

Instead of having the onIndexChanged be passed in via properties, add the current site to redux, with a reducer for changing state.

PR #23, opened 2024-09-29
Create useGenericStore

### LINKED ISSUE #9: Science state management
Currently, our state of the art state management system for science GUI is a custom hook I wrote in about 5 minutes called `useLocalStorage`. 

You use it identically to the `useState` hook, except you provide on additional argument which acts as the key to localStorage, and the state will be saved to localStorage.

This has architectural issues, wherein as data management requirements get more complicated, the hook gets passed down through many layers of the component tree, and the code gets terribly hard to read, manage, and maintain.

The benefit of using `useLocalStorage` was in the shear effectiveness of its simplicity -- it worked. It allowed me to cook, when all we needed was the feature to work the night before comp, rather than be "good" software.

We need to design a new state management system that can alleviate the scalability issues of `useLocalStorage` while also considering the readability and ease of use of the API through which you use the new state management system. 

Could you use redux?

See: 
- useLocalStorage
- Nir probe code
- UV vis spec code
- Thresholding code
- Redux docs

---

Some sources you may want to consider:
- https://www.npmjs.com/package/redux-state-sync
- https://www.npmjs.com/package/redux-persist 
- https://github.com/rt2zz/redux-persist-crosstab

Some problems you could try solve

### 1 -- **Kill** `useLocalStorage()`
Make an equally easy to use alternative, that doesn't result in horrible component architecture. There is a tendancy to create a god component that passes down the result of `useLocalStorage()` to everything that interacts with science state for some science instrument.

**Design a better persisted state convention**

This doesn't need to be specific to science, but you should create a better convention than `useLocalStorage()`;
- Persist
- Sync
  - mismatches in state could cause issues with persistence, where two diverging states try to overwrite each other!
  - could be between components, then even browser tabs
  - not really needed, but even between the two laptops in base station
- Be easy to use
  - i.e. I should never want to use `useLocalStorage()` when I have your API

### 2 -- Make it fit the needs of science

Often, science needs to get the data from your science instrument in a format they can use. 
For example, getting into a google doc of thiers, to create calibration curves from.

You should create a convention, that interfaces with your previous convention, for easily extracting this data from the gui for science to use.

Some ideas:
- export to csv, then upload to google docs or open in libre office
- copy to clipboard, paste in google doc

FelicityMatt:
## Description

Created useGenericStore to replace useLocalStorage. Basically a common and easy way to create redux stores which can then be easily used in a `[value, setValue]` way.

See [the doc I wrote](https://github.com/MonashNovaRover/nova/blob/7b64798df4f45cf712c7a5799134ce9a8a56debf/src/ros/nova-gui/docs/generic_store.md) for more detail.

fixes: #9 and helps #5 

Used `currentSite` as a way to test, will implement the SiteSelectWidget changes in a future PR.

## Changelogs

- create `createGenericStore`
- create `useGenericStore`
- when adding a new store default is to persist and tab sync -> the store must be manually blacklisted from both services if this functionality is not desired.

## Screenshots

[Screencast from 2024-09-30 09-12-49.webm](https://github.com/user-attachments/assets/593edc30-20c2-4cec-a502-0dfa14cbf6d9)

## Steps to Test

run the gui and check out the site selector and bottom counters. They both use `useGenericStore` and should persist and tab sync.

## Memes
useGenericStore


src/ros/nova-gui/nova-gui/src/redux/store/bifrost/bifrostStores.ts
```diff
@@ -0,0 +1,41 @@
+export const bifrostStores = [
```
FelicityMatt: Since I wanted to make it as easy as possible to onboard new stores this had to be created to blacklist all of the bifrost stores, the other option is to white list stores, which would greatly reduce the list, but would be an extra on boarding step for generic stores, so lmk what would be preferred.

src/ros/nova-gui/nova-gui/src/views/test/TestStateView/SiteSelectWidget.tsx
```diff
@@ -0,0 +1,94 @@
...
+    icon: (<Droplet/>)
+  },
+  {
+    type: SpaceResourceSiteType.ILMENITE,
+    name: "Ilmenite",
+    icon: (<Box/>)
+  }
+]
+
+const SiteSelectWidget: React.FC<SiteSelectWidgetProps> = (
```
FelicityMatt, line 35: Will move to this version of SiteSelectWidget in another pr, currently just a POC, not finalised yet

src/ros/nova-gui/nova-gui/src/redux/RootReducer.ts
```diff
@@ -223,4 +225,9 @@ export const rootReducer = combineReducers({
   ),
 
   cartographerState: cartographerSlice.reducer,
+
+  // generic stores
+  currentSite: createGenericStore("currentSite", 0),
+  currentSiteType: createGenericStore("currentSiteType", SpaceResourceSiteType.WATER),
```
FelicityMatt, line 235: not relevant to this PR but would we prefer site and site type to be two separate stores or combine them to be a single store with a struct? A single store seems cleaner and keeps the data together, but does involve a little more code for accessing and updating the values on the implementation side.

BaileyChessum: Current site type is just some extra metadata the NIR Probe associates with each site. Probably best not to make this super tightly coupled with the site selector itself, so we can keep it generic

src/ros/nova-gui/nova-gui/src/hooks/useLocalStorage.ts
BaileyChessum, line 1: Im going to cry

src/ros/nova-gui/nova-gui/src/redux/actions/useGenericStore.ts
```diff
@@ -0,0 +1,43 @@
+import {useDispatch, useSelector} from "react-redux";
+import {RootState} from "../RootState.ts";
+import {useCallback} from "react";
+import {GenericStoreState} from "../models/genericStores/GenericStoreState.ts";
+
+/**
+ * Like useState except uses a GenericStore in redux
+ *
+ * See the generic_store.md in the docs folder for more information
```
BaileyChessum: could this be a link to the doc? Fine if not tho

src/ros/nova-gui/nova-gui/src/redux/actions/useGenericStore.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const valueStore = store[storeName as keyof typeof store] as GenericStoreState<T>;
+
+    if (valueStore.value === undefined) {
+      console.error(`${storeName} is not a generic store`);
+      throw new Error(`${storeName} is not a generic store`);
+    }
+
+    return valueStore.value as T
+  })
```
BaileyChessum: This is a lot of logic for a useSelector, which could impact performance. Could this logic be placed anywhere else, with less calls, that would still achieve the same logic?

src/ros/nova-gui/nova-gui/src/redux/store/bifrost/bifrostStores.ts
```diff
@@ -0,0 +1,41 @@
...
+  "ramanSpecMessageStore",
+  "ramanMechMessageStore",
+  "ramanMechServiceStore",
+  "hydraprobeData",
+  "theta360CamStore",
+  "bmeSensorStore",
+
+  // Maps Related Stores
+  "roverLocationStore",
+  "baseLocationStore",
```
BaileyChessum: Could we automatically generate this from the RootReducer?

BaileyChessum (changes requested):
Some minor performance considerations to look into for the validation checks on the hook, and a developer experience thing with automatically blacklisting bifrost stores/actions

src/ros/nova-gui/nova-gui/src/redux/store/bifrost/bifrostStores.ts
FelicityMatt: I don't think we can since Root reducer is one big reducer and each field in it is a function so we can't check attributes.

src/ros/nova-gui/nova-gui/src/redux/store/rootReducerFilters.ts
```diff
@@ -0,0 +1,46 @@
+import {rootReducer, RootReducerKey} from "../RootReducer.ts";
+import {combineReducers} from "@reduxjs/toolkit";
+
+/**
+ * converts rootReducer which is a mix of StoreContexts and Reducers
+ * into a single reducer
+ */
+export const getReducers = () => {
```
BaileyChessum: Should root reducer be generated from the given store object once, or should we regenerate every time we want to use it? 

I feel like you should keep `rootReducer` with its original type definition, and just define it as `rootReducer = getReducers(theThingYouCurrentlyNamedRootReducer)`

```ts
// TODO: Give this a better, cooler name
const thingThatGeneratesTheRootReducer = {
  // ...
};

export const rootReducer = getReducers(thingThatGeneratesTheRootReducer);
```

src/ros/nova-gui/nova-gui/src/redux/store/configureRootStore.ts
```diff
@@ -1,40 +1,57 @@
...
+    }),
     // the function that is called for every action that matches one of the whitelisted actions.
     effect: tabSyncMiddleware(),
   };
 
   const listenerMiddleware = createListenerMiddleware();
 
   const store = configureStore({
-    reducer: persistReducer(persistConfig, rootReducer),
+    reducer: persistReducer(persistConfig, getReducers()),
```
BaileyChessum: see comment on function definition -- i think this line of code shouldn't have changed

src/ros/nova-gui/nova-gui/src/redux/store/rootReducerFilters.ts
```diff
@@ -0,0 +1,46 @@
...
+  const reducers = Object.keys(rootReducer).reduce((acc, val) => {
```
BaileyChessum: also, you should make the input to this function be a parameter, rather than referencing the thing you are currently calling `rootReducer` directly (it is no longer the rootReducer, but the thing that gets converted into the root reducer, and I don't think you should call it `rootReducer`)

Your code shouldn't be so dependent on our consuming code, it means it is super coupled to this one application, and can't be used anywhere else. Keep your code generic, reusable, and self-contained.

src/ros/nova-gui/nova-gui/src/redux/RootReducer.ts
```diff
@@ -15,9 +15,18 @@ import {uiSlice} from "./slices/UISlice";
...
-export const rootReducer = combineReducers({
+export type RootReducerKey = keyof typeof rootReducer;
+
+/**
+ * The rootReducer contains all values stored in redux.
+ *
+ * Each value in this struct can either be a Reducer or
+ * a StoreContext.
+ */
+export const rootReducer = {
```
BaileyChessum: This shouldn't be called the `rootReducer`, as it is no longer the root reducer. It's the input that is used to generate the root reducer.

Instead, I think you should make the `rootReducer` the original type

```ts
// TODO: Give this a better, cooler name
const thingThatGeneratesTheRootReducer = {
  // Bifrost Stores
  bifrostStatus: BifrostStatusStore(),
  
  // ...
};

export const rootReducer = getReducers(thingThatGeneratesTheRootReducer);
```

BaileyChessum (changes requested):
I really like this! I can't wait to start using this in GUI code. 

However, I have one minor gripe about the way you've written the `getReducers` function, which is a bit of a code smell. I think they way you use getReducers is weird. Why did you change `rootReducer` to be something different? Why not just make `getReducers` take the object as an input? Why does `getReducers` import the thing that generates the root reducer directly, rather than take it as an argument?

You should maybe think about getting in the habit of architecting your code for reusability, even if you aren't currently planning to reuse it. It simply encourages better and more maintainable designs. 

Think about your code as if you were making a library. Functions like `getReducers` should be independent on client code (i.e. the code specific to make our website). You should think about your code as if it is library code, that our client code consumes. It should be independent of logic specific to our website. 

Message me or arrange a time to call if you aren't sure exactly what I mean by this (particularly client code), and we can discuss it further.

Sorry if my comments seem pedantic, I am holding this to a slightly higher standard than normal, since this is effectively building the framework that our client code depends on. Client code can be sloppy, but I feel we should get the foundation right.

src/ros/nova-gui/nova-gui/src/redux/store/bifrost/bifrostStores.ts
BaileyChessum: this has been replaced by the new system for generating the `rootReducer`, right?

src/ros/nova-gui/nova-gui/src/redux/store/rootReducerFilters.ts
FelicityMatt: mm ok, i will fix this up, the only reason i left it as is was because i wanted to leave rootReducer unchanged so as the not confuse people who are familiar with the current setup

src/ros/nova-gui/nova-gui/src/redux/store/bifrost/bifrostStores.ts
FelicityMatt: yep this file does not exist anymore

src/ros/nova-gui/nova-gui/src/redux/actions/useGenericStore.ts
FelicityMatt: I wasn't able to link it to the file within the editor but have included a link to the doc in the master branch (which doesn't exist right now but will)

src/ros/nova-gui/nova-gui/src/redux/actions/useGenericStore.ts
FelicityMatt: moved to a useMemo which only runs on first render or when the storeName changes

src/ros/nova-gui/nova-gui/src/redux/store/bifrost/bifrostStores.ts
FelicityMatt: done

src/ros/nova-gui/nova-gui/src/redux/store/rootReducerFilters.ts
FelicityMatt: renamed to reduxStores

src/ros/nova-gui/nova-gui/src/redux/store/rootReducerFilters.ts
FelicityMatt: done, did the same with useGenericStore and the filter functions

src/ros/nova-gui/nova-gui/src/redux/RootReducer.ts
FelicityMatt: renamed to reduxStores and exported rootReducer at the bottom of the file

src/ros/nova-gui/nova-gui/src/redux/store/configureRootStore.ts
FelicityMatt: changed back

src/ros/nova-gui/nova-gui/src/hooks/useLocalStorage.ts
```diff
@@ -8,6 +8,8 @@ export type NoUnion<T, U = T> = T extends U ? [U] extends [T] ? T : never : neve
 export type NoConflict<T> = T & Exclude<T, undefined> & NoUnion<T>
 
 /**
+ * DEPRECATED please use useGenericStore instead
+ *
```
BaileyChessum: Mark this with jsdoc @deprecated: https://jsdoc.app/tags-deprecated

BaileyChessum (changes requested):

BaileyChessum (approved):
LGTM

PR #30, opened 2024-10-03
Fix drive controller compilation issues

BaileyChessum:
## Description

Simply changes `set_state()` to `set_lifecycle_state()` as per [this](https://control.ros.org/rolling/doc/ros2_control/doc/release_notes.html).


BaileyChessum:
I don't think this needs a review actually, changes are pretty trivial

PR #31, opened 2024-10-03
Feat/can speed increase

jchin14:
## Description

Increase default can0 speed to 250k
Update aliases from N1..3 to O1..3

## Steps to Test

Idk how to test on rover
Would just need to test this auto runs on startup and `candump can0` has an output
If change is successful, there will be an output. Nothing will show if unsuccessful

## Memes

the pics repo :(

### DIFF
--- a/nixfiles/modules/home/macros/can.sh
+++ b/nixfiles/modules/home/macros/can.sh
@@ -75,7 +75,12 @@ fi
 
 # If the bitrate parameter exists
 if [[ -z "${3:-}" ]]; then
-    bitrate=200000   # Default
+    if [[ $2 = "can0" ]]
+    then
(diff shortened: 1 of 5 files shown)


jchin14:
Pls squash before merging too

Redecorating:
I have simplified the systemd stuff and tested this on the rover and i can drive it with the xbox controller and candump can0 sends a lot of text. The other jetsons (ie not j1) will need to be moved to the monorepo for this to work on them, I am adding some instructions for that to the end of this page https://www.notion.so/Switching-to-the-mono-repo-5c1e7dca42324029bc10b5202f0e0af4

Redecorating (approved):

PR #33, opened 2024-10-04
Aliases for navigating to ~/nova and ~/nova/nixfiles

vic-bart:
## Description

Changed the `nova` alias to navigate to ~/nova instead of ~/nova/src, and added `nixfiles` alias to navigate to ~/nova/nixfiles.
Make sure to `sudo nixos-rebuild switch` and `reboot` for the aliases to take affect.

### DIFF
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -47,12 +47,12 @@ in
             visualisation = "${nova}/src/ros/visualisation";
             visualization = visualisation;
             vis = visualisation;
-            viz = vis;
+            viz = visualisation;
             science = "${nova}/src/ros/rover/science";
             camerasdir = "${nova}/src/ros/cameras2";
             autonomous = "${nova}/src/ros/rover/autonomous";
             auto = autonomous;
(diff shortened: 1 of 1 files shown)


nixfiles/modules/home/macros/default.nix
```diff
@@ -37,28 +37,29 @@ in
...
-            camerasdir = "${nova}/ros/cameras2";
-            autonomous = "${nova}/ros/rover/autonomous";
+            science = "${nova}/src/ros/rover/science";
+            camerasdir = "${nova}/src/ros/cameras2";
+            autonomous = "${nova}/src/ros/rover/autonomous";
             auto = autonomous;
-            gui = "${nova}/ros/gui";
-            tutorials = "${nova}/ros/tutorials";
-            pic = "${nova}/other/pics";
+            gui = "${nova}/src/ros/gui";
```
BaileyChessum: can we change this to nova-gui?

BaileyChessum (changes requested):

BaileyChessum (approved):
LGTM. Interesting to see so many directories that don't exist anymore

PR #80, opened 2024-10-05
Fix competition view button

### LINKED ISSUE #4: competition view drop down doesn't match competition view
## Description
The top right drop down for selecting which competition routes/view to be in does not change when a different competition is selected, it always says `general` which is not accurate.

## Screenshot
eg here it says `general` even after selecting `arc` from the drop down

Image

FelicityMatt:
## Description

The top right competition view button now reflects the current page you are on.

fixes: #4 

## Changelogs

- button content dynamically changes with the view.

## Screenshots
image

## Steps to Test

- change views!

## Memes

image

### DIFF
--- a/src/ros/nova-gui/nova-gui/src/components/TopBar/TopBar.tsx
+++ b/src/ros/nova-gui/nova-gui/src/components/TopBar/TopBar.tsx
@@ -21,6 +21,7 @@ import humanizeString from "humanize-string";
 import { BLCMDStatusButton } from "../BLCMDStatusModal/BLCMDStatusButton";
 import "./TopBar.css";
 import { List } from "react-bootstrap-icons";
+import {useState} from "react";
 
 const connectionStatusColor: {
   [key: string]: "success" | "warning" | "danger";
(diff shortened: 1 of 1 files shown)


BaileyChessum:
image

src/ros/nova-gui/nova-gui/src/components/TopBar/TopBar.tsx
```diff
@@ -98,12 +106,11 @@ export const NovaTopBar: React.FC = () => {
...
             <DropdownTrigger>
-              <Button
-                radius="sm"
-                size="sm"
-                endContent={<ChevronDown className="w-4 h-4" />}
-              >
-                General
+              <Button radius="sm" size="sm">
+                <div className="w-10">
+                  {prettyViewNames.has(viewName) ? prettyViewNames.get(viewName) : "General"}
```
BaileyChessum: Should the fallback name indicate that there is an error of some kind? Should it be "unknown" or something similar? 

Could get confusing unless the side navigation also shows general views when the view name is unknown.

BaileyChessum (changes requested):

src/ros/nova-gui/nova-gui/src/components/TopBar/TopBar.tsx
FelicityMatt: changed to ???

BaileyChessum (approved):
LGTM

PR #81, opened 2024-10-05
Encrypt secrets using sops-nix

vic-bart:
## Description

This PR was made by @dylan-gonzalez on the nixfiles repo.

Our secrets are currently not encrypted in our repo. This PR implements [sops-nix](https://github.com/Mic92/sops-nix) to encrypt relevant secrets.

I have not properly tested the CI module, only that I've checked that the secrets are being evaluated correctly. I might get time later in the week to test, otherwise I'd appreciate if someone else could.

The secrets that are now managed with sops-nix are:

the password to access the hydra binary cache - hydra/password
the hydra git private ssh key - hydra/git_ssh_key
the hydra binary cache secret - hydra/binary_cache_secret
the hydra github token - hydra/github_token
Note: this has only been implemented as a nixos module, so it will not work with home-manager at this stage.


vic-bart:
Trying to get nova/hashedPassword building, so I added this:

```nix
    secrets = lib.mkMerge [
      {
        "nova/hashed_password" = {};
      }
```

Which lets it build, but still throws the following error:

image

And now the normal password for logging into nova / using sudo commands no longer works.

nixfiles/modules/nixos/common/users/default.nix
dylan-gonzalez, line 1: @vic-bart I must have left this here by accident.. try going back to the original

vic-bart:
TL;DR, Bailey tried this and Hydra had issues, so this development is being shelved for the time being

PR #92, opened 2024-10-13
Fix camera.launch.py

vic-bart:
## Description

Initially just to fix a few minor changes causing the oak nodes to fail, @hacker1024 pointed out I shouldn't be using absolute filepaths based off my own file system structure. 

These filepaths are used in parameters for the oak camera (depthai_oakd_rgbd.yaml) and the camera's nn model (.json files), and need to be absolute (relative filepaths would be independent of file system structure, but would require users to always launch from a specific directory, which sucks). 

Because YAML and JSON (as far as I can tell) don't support any way to natively generate absolute filepaths, and filepaths can't be generated after being built into the nix store (nix store is a read-only file system), I've made a change during the postInstall phase of auto_bringup to manually overwrite those values with absolute paths dependent on the nix store hash (which will always be present).

## Steps to Test

- `ws-build`
- Check the parameter values for the relevant yaml and json files in `result/share/auto_bringup`
- Enter `result/bin/ros2 launch auto_bringup camera.launch.py` and make sure it works (no errors or unexpected behaviour)

## Memes

image

### DIFF
--- a/src/ros/rover/auto_bringup/CMakeLists.txt
+++ b/src/ros/rover/auto_bringup/CMakeLists.txt
@@ -25,6 +25,7 @@ find_package(ament_cmake REQUIRED)
 install(DIRECTORY
   launch
   params
+  resources
   rviz
   DESTINATION share/${PROJECT_NAME}
 )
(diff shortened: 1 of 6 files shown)


hacker1024 (changes requested):
There should be no assumptions made about filesystem structure outside of this repo.

vic-bart:
Tested by @tarikthomas, approved by @KABILAN235.

PR #102, opened 2024-10-20
Migrate auto_bringup to Gazebo Harmonic superficially

BaileyChessum:
## Description
Description

Follows [this guide](https://gazebosim.org/docs/harmonic/migrating_gazebo_classic_ros2_packages/) to migrate auto_bringup from Gazebo Classic to Gazebo Harmonic

WIP because running gazebo through ros2 is broken, where it runs ruby $(which gz) sim-1. The person who made the PR we are using currently said they changed their launch file to avoid using ruby $(which gz) sim-1, but didn't fix it. We either need to fix it, or do the same to the launch files.

### DIFF
--- a/src/ros/rover/auto_bringup/launch/gazebo.launch.py
+++ b/src/ros/rover/auto_bringup/launch/gazebo.launch.py
@@ -13,11 +13,12 @@ PACKAGE: 	core
 CREATION:	27/04/2023
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 """
+import os
 
 from ament_index_python.packages import get_package_share_path, get_package_share_directory
 
 from launch import LaunchDescription
-from launch.actions import DeclareLaunchArgument, IncludeLaunchDescription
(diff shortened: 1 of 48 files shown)


src/ros/rover/auto_bringup/launch/gazebo.launch.py
```diff
@@ -86,12 +90,13 @@ def generate_launch_description():
     )
 
     start_gazebo_server_cmd = IncludeLaunchDescription(
-        PythonLaunchDescriptionSource(PathJoinSubstitution([gazebo_dir, 'launch', 'gzserver.launch.py'])),
-        launch_arguments={"world": world}.items()
+        PythonLaunchDescriptionSource(os.path.join(ros_gz_sim, 'launch', 'gz_sim.launch.py')),
```
vic-bart: Why the switch from `PathJoinSubstitution` to `os.path.join`? They do the same thing, but I'd prefer to use an existing package then import a new one exclusively for this

vic-bart: fixed this

vic-bart:
Checked over, looks good

PR #118, opened 2024-11-20
Merge Fix/nav2 rviz plugins

### LINKED ISSUE #96: figure out why rviz nav2 plugins brokey
**CONTEXT**
The migration to Gazebo Harmonic has broken the nav2 plugins for rviz. Why? That's for you to find out!

**TODO**
- Identify which rviz nav2 plugins broke
- Identify why they broke

**RESOURCES**

V01DBREAKER:
Fixes the nav2_rviz_plugins so that they are no longer brokey.

## Description

This removes all mention of the ` updateAutoDeactivate()` function from the c++ code as it was causing issues loading the plugins in rviz.

details here: https://www.notion.so/Figure-out-why-rviz-nav2-plugins-are-brokey-3f104d21bc9c4881ba702ab44a1cbfdf

## Changelogs

maintanence.nix has some added code to modify the postPatch phase of building the package.

## Screenshots

image

## Steps to Test

run `ros2 launch auto_bringup rviz.launch.py` and enjoy the now working nav2 panel 
or run `ros2 run rviz2 rviz2` and attempt to use any of the nav2 plugins 

## Memes
[Me when i copy rohit's moveit2 nix code](https://www.youtube.com/shorts/H3MgcNJG8J8)

Thanks Rohit for teaching me how to patch with nix!

### DIFF
--- a/nixfiles/overlay/ros/maintanence.nix
+++ b/nixfiles/overlay/ros/maintanence.nix
@@ -148,6 +148,7 @@ self: super:
 
       nav2-rviz-plugins = rosSuper.nav2-rviz-plugins.overrideAttrs ({ postPatch ? "", ... }: {
         # remove broken updateAutoDeactivate() symbol, 20/11/2024, Navigation2 1.3.2
+        # https://github.com/MonashNovaRover/nova/issues/96
         postPatch = postPatch + ''
           substituteInPlace src/costmap_cost_tool.cpp --replace "SLOT(updateAutoDeactivate())" "nullptr"
           substituteInPlace include/nav2_rviz_plugins/costmap_cost_tool.hpp --replace "private Q_SLOTS:" "" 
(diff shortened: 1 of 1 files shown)


RohitPilakkat:
for more context, updateAutoDetection is only being used in this particular tool (And specifically for the properties panel functionality) and has been defined as a private Q_Slot but has not been given any functionality (And does not seem to have any when searching nav2 repo) so it exists but is not defined or usable.

No idea what the original purpose of this method was, but the easiest solution was just to remove references to it and the symbolic definition of it as a private Q_Slot

vic-bart:
Looks great, I'll close with the caveat that there is a possibility this change **may** introduce anomalous behaviour to how nav2 rviz2 plugins function, however it's not feasible to thoroughly test this with the current state of nav2 integration. Should this possibility occur, it would be detached enough from this development to warrant a separate issue and PR; this PR has accomplished what it set out to do &mdash; figure out why rviz nav2 plugins brokey.

PR #121, opened 2024-11-28
Extending Gazebo Harmonic functionality

### LINKED ISSUE #95: identify ROS topics needed from gazebo
**CONTEXT**
Gazebo Classic plugins published to ROS topics by default, but Gazebo Harmonic plugins do not &mdash; instead publishing to "Gazebo topics" on an internal Gazebo network that are not exposed to the ROS network. 
For these topics to be published as ROS topics, a "bridge" that maps Gazebo topics to ROS topics needs to be created using "ros_gz_bridge" (see resources for more).

**TODO**
- Join #gazebo-harmonic on slack (if you haven't already)
- Identify the ROS topics that need to be published from Gazebo

**RESOURCES**
- [gz integration with ros via ros_gz_bridge](https://gazebosim.org/docs/latest/ros2_integration/)

vic-bart:
## Description

This PR introduces a list of new functionality for Gazebo Harmonic
- ros_gz_bridge &mdash; Gazebo plugins do not work natively with ROS2 anymore, only publishing to Gazebo topics, which now need to be republished on the ROS network via ros_gz_bridge. There are still some minor issues to be resolved with certain plugins, but these will be addressed in future PRs.
- urdfs &mdash; There were several files in rover_description that were leftovers from when the rover urdf was exported, or otherwise no longer being used. These have been removed. Changes to the rover urdf include removing all collision meshes except for the wheels to improve performance.
- launch files &mdash; gazebo, control, and urdf.launch.py were refactored to remove redundancies and be easier to understand. This has highlighted the need to rework and create new launch files, which will be addressed in a future PR.
- worlds &mdash; minor changes to the worlds in nova_gazebo (plus a new one from [gz-sim/examples/worlds](https://github.com/gazebosim/gz-sim/blob/gz-sim8/examples/worlds/empty.sdf)), with further changes to be developed in a future PR.

## Steps to Test

- `ws-build`
- `result/bin/ros2 launch auto_bringup gazebo.launch.py`

## Memes

image


RohitPilakkat:
Won't removing collisions from the body cause future issues with path planning such as with the arm? Especially if it's using moveit2 to path plan.

I think a better way to do this would be to do some pure shape approximations using OpenSCAD or similar

BaileyChessum (approved):
Didn't get the opportunity to review this too thoroughly, but from what I can tell it seems good

BaileyChessum:
> Won't removing collisions from the body cause future issues with path planning such as with the arm? Especially if it's using moveit2 to path plan.
> 
> I think a better way to do this would be to do some pure shape approximations using OpenSCAD or similar

yes, but I think we should worry about that when we make the rover 7 urdf

PR #125, opened 2024-12-10
Added some rover aliases to default.nix

jaredlandau:
Added a rover-help command, as well as launch-base and launch-drive aliases that will help minimise the number of lengthy commands and arguments people need to memorise


vic-bart:
works

PR #126, opened 2024-12-12
fix: nix-channels update fixes

BaileyChessum:
## Description

Nerdfonts is broken and preventing me from building. So, I commented it out. This PR should be followed by an attempt to actually understand what it is and replace it with a newer equivalent.

## Steps to Test

Update your nix channels, then rebuild


RohitPilakkat:
made some changes that seem to work fine for me.
also added in a small fix for desktop since it's related to similar issues.

BaileyChessum (commented):
Looks good @RohitPilakkat -- I cannot approve this since I made it though

vic-bart:
works

PR #129, opened 2024-12-12
New Rover URDF with suspension

### LINKED ISSUE #48: Create rover URDF for Gazebo Harmonic
We want to move to Gazebo Harmonic, which doesn't allow for closed kinematic chains, only tree like chained structures.

1. Create a URDF of the rover, without implementation of the diff bar (see nova knowledge base > dictionary)
2. Research alternatives to closed kinematic chains to model the behaviour of the diff

[OnShape-to-URDF/SDF](https://onshape-to-robot.readthedocs.io/en/latest/)
Tried this, didn't work
[SolidWorks-to-URDF](https://wiki.ros.org/sw_urdf_exporter)
What we've used before; could be used to create URDF and converted to SDF by hand (very tedious)

### LINKED ISSUE #55: improve rover suspension and model in gazebo
**CONTEXT**
The rover's suspensions in gazebo is currently static (which isn't very accurate), and should be changed to dynamic (which would require closed kinematic chains, CKCs). 
The current file format defining the rover model in gazebo, URDF, doesn't support CKCs, but another file format, SDF, does (interesting fact: gazebo actually works with SDFs under-the-hood; URDFs are converted into SDFs before being used by gazebo).
The rover itself has changed significantly since the original URDF was made, so this would also entail a redesign of the rover's entire model.
- [SDF guide](https://sdformat.org/)
- [OnShape-to-URDF/SDF](https://onshape-to-robot.readthedocs.io/en/latest/)
  - Tried this, didn't work
- [SolidWorks-to-URDF](https://wiki.ros.org/sw_urdf_exporter)
  - What we've used before; could be used to create URDF and converted to SDF by hand (very tedious)

**TODO**
- Get up-to-date rover model from OnShape
- Export to SDF, or URDF (then convert to SDF)
- Fine tune SDF parameters to have accurate rover suspension

vic-bart:
## Description

A new rover URDF based on our latest rover 7!
Generated using the onshape-to-robot tool.

The tools to generate the URDF available using `urdf-tool` which creates a nix-shell with onshape-to-robot, openscad and meshlabs. (Thanks Rohit!)

An example config.json to use with the tool as well as the `.scad` files used to make the pure shape approximations for collisions.

A lot of the process is documented here:
https://www.notion.so/URDF-or-SDF-Generation-Using-Onshape-to-Robot-bb5aeae26f7644438dcc936960faecb3
Addresses #48 and #55

## Changelogs

Differs from the old rover urdf in that:
* Updates the model to the latest rover 7!
    * Implements differential bar, including the closed chain kinematic loops using the detachable joint plugin to close the loop.
    * Updates the controller parameters to match the new model
    * Updates the launch files with a new parameter for the rover name (needed for the plugin)
* Adds a nix shell file that can be run to get the tools for to make another urdf
* Updates the file structure of rover_description to separate the arm from the rest of the urdf

## Screenshots

image
image

https://github.com/user-attachments/assets/733d597e-f142-46dd-9c90-5a560da814b2

## ROS

Overrides the existing rover_description package

## Steps to Test

Run `ros2 launch auto_bringup everything.launch.py` or `ros2 launch auto_bringup gazebo.launch.py`
To use the tools, run `urdf-tool`. (READ THE README in `rover_description`)

## Memes

Learnt how to use Onshape, how URDF and SDF works. How gazebo works, how launch files work, etc. - Anthony

image


V01DBREAKER:
Due to changes to the URDF the ros joint state broadcaster node didn't broadcast any states past the leg joints (since they are dynamic joints and i assume not in our config). I couldn't figure out how to solve, this and instead used the joint broadcaster gazebo plugin (different from the ros node). This may affect the TF in rviz on the actual rover.

~~Also i have not touched nova_bringup at all, and things may be affected there due to file changes.~~ 
should be solved in the following two commits

V01DBREAKER:
Besides the merge conflicts, the gz_attach_links module is no longer needed and can be removed @vic-bart (unless u still need ot for something im going to remove it).

I can't resolve them which is annoying but basically the gazebo.launch.py changes are fine as long as the rover name argument is still passed to the urdf.launch.py (while z pose of rover = 0)
And the old urdf is deleted anyway

vic-bart (commented):
Very good work! We got a new rover :sunglasses:

PR #133, opened 2024-12-13
cd nova, cd nixfiles, cd rover

vic-bart:
## Description

We're cding places.
- Removed old alias
- Added support for `cd nova/nixfiles/rover/science/gui`, while still keeping their alias equivalents

## Steps to Test

- `sudo nixos-rebuild switch`
- Try them out!

## Memes

image

### DIFF
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -25,18 +25,18 @@ in
       complete -F _complete_alias "''${!BASH_ALIASES[@]}"
       
       cd() {
-        if [ ! -d "$1" ] && [ "$1" = "nova" ]; then
-          builtin cd "${cfg.sourceDir}/.."
-        elif [ ! -d "$1" ] && [ "$1" = "nixfiles" ]; then
-          builtin cd "${cfg.nixfileDir}"
-        elif [ ! -d "$1" ] && [ "$1" = "rover" ]; then
-          builtin cd "${cfg.sourceDir}/ros/rover"
-        elif [ ! -d "$1" ] && [ "$1" = "science" ]; then
-          builtin cd "${cfg.sourceDir}/ros/rover/science"
-        elif [ ! -d "$1" ] && [ "$1" = "gui" ]; then
-          builtin cd "${cfg.sourceDir}/ros/nova-gui"
-        else
-          builtin cd "$1"
+        if [ -d $1 ]; then
+          builtin cd $1
+        elif [ $1 = nova ]; then
+          builtin cd ${cfg.sourceDir}/..
+        elif [ $1 = nixfiles ]; then
+          builtin cd ${cfg.nixfileDir}
+        elif [ $1 = rover ]; then
+          builtin cd ${cfg.sourceDir}/ros/rover
+        elif [ $1 = science ]; then
+          builtin cd ${cfg.sourceDir}/ros/rover/science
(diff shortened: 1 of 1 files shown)


hacker1024:
I don't agree with this interface.

It is very counterintuitive to add fake targets to `cd` like this, especially if one ever needs to change into a real local directory with one of these names, and especially if they don't know the alias exists in the first place.

With a pure alias, it is obvious that a custom command with a non-trivial implementation is being run.

If you really want to add directory aliases that can be used in commands, I recommend [Zsh's functionality](https://til.hashrocket.com/posts/xsavbhlrz4-shortcuts-with-hash-d-in-zsh).

nixfiles/modules/home/macros/default.nix
```diff
@@ -23,6 +23,22 @@ in
...
+        elif [ "$1" = "rover" ]; then
+          builtin cd ~/nova/src/ros/rover
+        elif [ "$1" = "science" ]; then
+          builtin cd ~/nova/src/ros/rover/science
+        elif [ "$1" = "gui" ]; then
+          builtin cd ~/nova/src/ros/nova-gui
+        else
+          builtin cd "$@"
+        fi
+      }
```
hacker1024: You cannot assume that the user keeps Nova source files in the root of their home directory. 

There are NixOS options that should be used instead with string interpolation. See the other aliases for examples.

hacker1024 (changes requested):

nixfiles/modules/home/macros/default.nix
vic-bart: I've swapped them out to use the string interpolation.

vic-bart:
Am I correct in saying that implementing [Zsh's functionality](https://til.hashrocket.com/posts/xsavbhlrz4-shortcuts-with-hash-d-in-zsh) would mean commands look like `cd ~nova`? In which case, we might as well leave it as `nova`. The purpose of this PR was to specifically allow for `cd nova` (and other aliases) to have the same functionality as `nova` (and their equivalents) &mdash; see [Slack thread](https://monashnovarover.slack.com/archives/C07H7FA2A3Z/p1733540167669469) for context &mdash; so changing it would introduce further confusion.

I've amended the bash shell function to only overwrite the `cd` command if the filepath doesn't exist. For example, if your directory is /home and you want to go to /home/nova, `cd nova` will work as intended. If `nova` isn't a valid filepath, then it will default to `cd ~/nova/src/..`.

Are there any problems with this; in what scenario would this cause issues?

[edit] wait something broke 
[edit] [fixed it](https://stackoverflow.com/questions/12133209/bin-bash-cd-results-in-bin-bash-cd-no-such-file-or-directory-why), "" suck

vic-bart:
I've since realised that one can overwrite their `cd` within their own NixOS; there's no need to enforce this on everyone, I'll just pin a tutorial somewhere that shows how to implement it in their own NixOS if people want.

RohitPilakkat:
> I've since realised that one can overwrite their `cd` within their own NixOS; there's no need to enforce this on everyone, I'll just pin a tutorial somewhere that shows how to implement it in their own NixOS if people want.

There's a snippet on Notion that I made ages ago that covers how to add and edit aliases on your own system fyi

https://www.notion.so/NixOS-Overview-Customisation-961daf112a4e4d51b5ff681dc40d64dd?pvs=4#2bb03321b6694bf8b747df643644a071

PR #138, opened 2024-12-16
ARCh gazebo world

vic-bart:


vic-bart:
works, and to clarify the borked commit history, @JoelAKruger was the mastermind behind developing this code, I just reviewed it!

PR #143, opened 2024-12-19
Add GUI build workflow

### LINKED ISSUE #11: Simple CI/CD pipeline for tsc warnings
Warnings in the nova-gui build prevent nixos from building. These are usually hard to catch in the code review process, and can become a headache to those trying to build nixos.

Here, for example, some unused variables prevent nixos from rebuilding:

Image

- Add a github action to check for these warnings in PRs
- Maybe use something like [this](https://github.com/marketplace/actions/run-tsc-with-reviewdog) to show the warnings in the PR discussion

FelicityMatt:
## Description

Basic GitHub workflow that builds the nova-gui and will fail if there are any build issues.

Only caviat is that despite using the Nix Magic Cache action, the build still takes **an hour**, but I figure this is a last resort check, and most PRs will be waiting for review/worked on so shouldn't been an issue, but obviously not ideal. I will create a follow up issue to be put in the backlog.

fixes #11 

## Changelogs

- add a GitHub workflow to build the gui

## Screenshots

[Workflow](https://github.com/MonashNovaRover/nova/actions/runs/12407946081) failed when there was a build error in the gui:

Screenshot From 2024-12-19 20-39-27

image

[Workflow](https://github.com/MonashNovaRover/nova/actions/runs/12407222047/job/34636904630) passes when there were no build errors:

Screenshot From 2024-12-19 20-39-51

## Steps to Test

- commit a change that causes a build error and create a PR then wait an hour.

## Memes

image

### DIFF
--- a/.github/workflows/gui-builder.yml
+++ b/.github/workflows/gui-builder.yml
@@ -19,4 +19,4 @@ jobs:
       - uses: DeterminateSystems/nix-installer-action@main
       - uses: DeterminateSystems/magic-nix-cache-action@main
       - name: Run `nova-build`
-        run: nova-build -A pkgs.ros.nova-gui
+        run: nix-build nixfiles -A pkgs.ros.nova-gui
--- a/.github/workflows/gui-builder.yml
+++ b/.github/workflows/gui-builder.yml
@@ -1,15 +1,17 @@
+# Follows template from: https://github.com/DeterminateSystems/magic-nix-cache-action/
+
 name: gui builder 🔨
 
 on:
   pull_request:
-    # paths:
-    #   - 'src/ros/nova-gui/nova-gui/**'  # Only run when files in the nova-gui folder change
+    paths:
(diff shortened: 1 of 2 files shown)


rotaryViper:
Might be worth checking out this file for an alternate way of seeing how to build nova-gui, using yarn install instead of nova-build.

FelicityMatt:
> Might be worth checking out this file for an alternate way of seeing how to build nova-gui, using yarn install instead of nova-build.

We've tried using yarn build before but we ran into the issue that the rosTypes file doesn't exist locally, instead it's a symlink to another file which is defined when nova-build is run.

BaileyChessum:
@rotaryViper I think the only thing preventing us from doing this is that we depend on ros-typescript-generator to generate message types in the nix build process, so it would be missing a file if we just did yarn directly. 

@FelicityMatt could we manually invoke ros-typescript-generator to avoid this?

BaileyChessum:
I'm happy to merge this in for now after creating a follow up (low priority) issue to reduce the build time

BaileyChessum (approved):

PR #152, opened 2025-01-03
Reviving waratah and fixing nav2

vic-bart:
## Description

- Revived waratah model urdf so it can still be used while rover 7 is assembled
- Updated nav2 params and patched robot_localization to use TwistStamped msgs instead of Twist msgs (which the whole ROS2/Nav2 ecosystem is moving towards, and a change we've already implemented for our controllers &mdash; so technically, this is a bug fix :flushed:)
- Changed ros2 controller params for auto_bringup to be in autonomous mode by default (to prevent this being problematic for manual operators in the future, nova_bringup will have a separate params file for teleop)

## Steps to Test

- `ros2 launch auto_bringup everything.launch.py` and drive around with BT
- `ros2 launch auto_bringup everything.launch.py model:=<filepath to rover_description/waratah/urdf/rover.urdf.xacro> controllers:=<filepath to auto_bringup/params/old_controllers.yaml>` and drive around with BT

## Memes

image


vic-bart:
I've tested this in sim and on rover, works as intended without errors, so just need an extra set of eyes :eyes: to review the code changes

KABILAN235 (changes requested):
I think it would be cool to move the rover7's URDF into a `rover7` directory like waratah, would make it more explicit that these are 2 different versions of the rover

vic-bart:
done

KABILAN235 (approved):
**L**ets **G**et **T**his **M**erged

Issue #46, opened 2024-09-08
ros2_control NIR probe


BaileyChessum:
Convert the NIR probe (`nir_probe_publisher.py`) to use ros2_control

Make hardware interfaces

[ros2_control hardware interface types -- sensor](https://control.ros.org/master/doc/ros2_control/hardware_interface/doc/hardware_interface_types_userdoc.html#sensors)

NIR probe inputs:
- current LED : byte (0 = off, 1 = on)
- current value : int

NIR probe outputs:
- set current LED : byte (0 = off, 1 = on)

You'll then need to make a broadcaster:

[ros2_controllers -- Broadcasters](https://control.ros.org/master/doc/ros2_controllers/doc/controllers_index.html#broadcasters)

FelicityMatt:
![Image](https://github.com/user-attachments/assets/130ae2aa-a671-4c93-84e3-5507cdbae62b)

TJ-Lee-M:
See #166 for ros2 (not ros2_control) implementation for the 24/25 design cycle

V01DBREAKER:
Plan is not to use ros2_control for science.
WE LOVE SLOP!!!

PR #166, opened 2025-01-10
[WIP] Updated ros2 nir for new probe

### LINKED ISSUE #171: Update NIR Probe ROS2 Publisher
The current NIR Probe ROS2 publisher needs to be updated to support two LEDs and Photodiodes, as well as automatic readings.

rotaryViper:
## Description

This is the python replacement for the long awaited nir probe code. (See #46 for original intention with ros2_control)
I can't run it because I don't know how to mock the nir probe. Just believe.
10 weeks of ros2 control done in 10 minutes of python.

It supports both lighting up either nir probe or turning off both at the same time.
It sends photodiode data to seperate CAN IDs.

## Changelogs

Supports 1 new light and 1 new photodiode.
Slightly improved error handling for handling can IDs.

## ROS
This should probably be rewritten in ros 2 control by someone smarter than me.

## Steps to Test
🤷

## Memes
idk thinking about the meme took longer than writing the code


rotaryViper:
Its done btw, unless if i'm supposed to update a xacro or something

rotaryViper:
to do: investigate filter masking for receiving can ids

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -30,16 +30,21 @@ class NIRProbePublisher(Node):
     CAN_BUS = "can1"
     # Card IDs
     NIR_PROBE_ID = 0x0F0
-    CARD_ID_RECEIVE = 0x4F1
+    CARD_ID_RECEIVE_PD1 = 0x4F0
```
FelicityMatt: is it worth adding a small comment saying PD = Photodiode or receive readings because I can see future people being very confused.

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -30,16 +30,21 @@ class NIRProbePublisher(Node):
...
+    CARD_ID_RECEIVE_PD2 = 0x4F1
+
     # Command data
-    NIR_PROBE_LED_ON = 0x01
-    NIR_PROBE_LED_OFF = 0x02
-    NIR_PROBE_READ = 0x03
+    NIR_PROBE_LED1_ON = 0x01 # Also turn LED2 off
+    NIR_PROBE_LED2_ON = 0x02 # Also turn LED1 off
+    NIR_PROBE_LED_OFF = 0x03
+    NIR_PROBE_READ_P1 = 0x04
```
FelicityMatt: Maybe rename to `NIR_PROBE_READ_PD1` to be consistant with  `CARD_ID_RECEIVE_PD1`

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -71,50 +80,89 @@ def send_read_command_callback(self):
...
+        """
+        Callback for when data is received from the NIR probe
+        """
+        self.get_logger().debug(f"Received {hex(frame.id)} {frame.data}")
+
+        # Receive data if corresponding LED is on
+        match frame.id:
+
+            case self.CARD_ID_RECEIVE_PD1:
+                if self.led == self.LED1_BYTES_ON:
```
FelicityMatt: question: do we want to publish data only from the LED we've switched on or from both?
and if we only publish the one we've switched on is requesting data from both LED's then create unused/unnecessary requests/can bus use?

rotaryViper:
The code i wrote was bad, so i rewrote it

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +35,165 @@
...
+            self.get_logger().debug(f"Sending {led_frame}")
+            self.bus.send(led_frame)
             response.success = True
+
         except Exception as e:
             self.get_logger().error(f"Failed to send led command over CAN: {e}")
-            response.success = False
 
+        self.led = request.led
         return response
```
FelicityMatt, line 171: would it be possible to turn on/off the LED and return a reading here instead of a success variable?

rotaryViper: We need to return the response True/ False because its a service that expects a callback. Response is declared in the nova_interfaces folder. But we can add another message to log that the NIR probe has sent the command successfully

FelicityMatt: can we change the service definition to return a value? we don't currently use the true/false response in the GUI

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
+    ]
+
+    # Timings
+    SEND_INTERVAL                 = 0.01
+    READ_INTERVAL                 = SEND_INTERVAL
+
+    # Number of readings before moving on
+    PHOTODIODE_LIGHT_BLANK_PERIOD = READ_INTERVAL * 100
+    PHOTODIODE_CALIBRATION_PERIOD = READ_INTERVAL * 100
+    PHOTODIODE_ONLINE_PERIOD      = READ_INTERVAL * 30
```
TJ-Lee-M: number of readings for each stage currently arbitrary place holder numbers - actual number to be chosen with science

src/ros/rover/nova_interfaces/msg/NIRProbeData.msg
```diff
@@ -2,12 +2,19 @@
...
+byte LED_ICE = 2
 
 # The LED (if any) used to take the reading
 byte led
+
+# Determine on GUI if LED is on
+bool led_on
+
+# Determine on GUI if photodiode is on
+bool photodiode_on
```
FelicityMatt: Upon first look these new fields are a bit confusing and I'm unsure the significance of them, whats the difference between led and led_on are they not the same? and is the photodiode not always on?? I will come back after reading the code, but I feel this should be super clear so that one doesn't have to look at the code to figure out the significance.

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -4,21 +4,30 @@
...
+CREATION:    14/01/2025
+EDITED:      15/01/2025
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+This package bridges turning on LEDs with
+collecting NIR photodiode data.
+
+The only commands to send this publisher are:
+    0. Turn off LEDs
+    1. Turn on LED 1
+    2. Turn on LED 2
```
FelicityMatt: will need to update this with finalised service

src/ros/rover/science/science/nir_probe_publisher.py
FelicityMatt: Just clarifying this indicates the amount of time that it reads light blank readings, then waits for a calibration period then reads data for a certain period? Might want to make a clear statement about this process in the file header and ensure all terms are clear and clarified.

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
+        # Only accept inputs being delivered to receiving CAN IDs
+        self.bus.set_id_filter(self.NIR_RECEIVE_ID)
+        self.bus.add_callback(self.PHOTODIODE1_ID, self.read_data_callback)
+        self.bus.add_callback(self.PHOTODIODE2_ID, self.read_data_callback)
 
-        self.publisher = self.create_publisher(NIRProbeData, '/science/nir_probe_data', 10)
+        self.bus.open(self.get_parameter(self.CAN_BUS_PARAM).value)
 
+        # Initialise service for taking commands
         self.led_service = self.create_service(SetNIRProbeLED, '/science/set_nir_probe_led', self.led_service_callback)
```
FelicityMatt: I think we should rename this service, `set_led` doesn't really fit anymore

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
+
+            match self.led:
+                case self.LED1_ON:
+                    command = self.READ_PHOTODIODE1
+                case self.LED2_ON:
+                    command = self.READ_PHOTODIODE2
+                case self.LED_OFF:
+                    return
+                case _:
+                    self.get_logger().error(f"Invalid LED request received by read data callback: {self.led}")
```
FelicityMatt: just confirming we don't want to return here and not send a can command here if there's an issue?

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
-        Sends the read command to the NIR probe
+        Sends the read data command to the NIR probe
         """
-        frame = jcan.Frame(self.NIR_PROBE_ID, [self.NIR_PROBE_READ])
-        
         try:
-            self.bus.send(frame)
+            command = self.TURN_LED_OFF
+
+            match self.led:
```
FelicityMatt: super nit picky but do we need this switch statement within the try? I think its better practice to only put necessary code within the try.

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
-        else:
-            self.get_logger().info("Turning NIR probe LED ON")
-            self.led = self.LED_BYTES_ON
-            frame = jcan.Frame(self.NIR_PROBE_ID, [self.NIR_PROBE_LED_ON])
+        response.success = False
+
+        message = int.from_bytes(request.led, "big")
+
+        # Check which led state NIR Probe publisher is in
+        match request.led:
```
FelicityMatt: have you considered what will happen here if this service gets called while another automatic reading is taking place?
You should write this approach in a comment somewhere.

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
 
+        response.success = True
+
+        self.led = request.led
+
+        self.get_logger().debug(f"Activating photodiode {request.led}")
+
+        self.publish_msg(-1, -1, False, True) # -1 to not consider this value
+
+        return response
```
FelicityMatt: honestly I still believe that the "most correct" way to use services is to return the value that we are seeking within the service message/topic, but if you believe that this method would work better that's probably ok.

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
+
+        self.frame = jcan.Frame(self.NIR_PROBE_ID, [message]) # Save frame to turn on light at the end of light blanking
 
+        response.success = True
+
+        self.led = request.led
+
+        self.get_logger().debug(f"Activating photodiode {request.led}")
+
+        self.publish_msg(-1, -1, False, True) # -1 to not consider this value
```
FelicityMatt: what does this mean? pls explain the sequence of published messages somewhere

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
-        self.bus.open(self.get_parameter(self.CAN_BUS_PARAM).value)
 
-        self.timer = self.create_timer(0.1, self.send_read_command_callback)
-        self.timer_jcan_spin = self.create_timer(0.01, self.bus.spin)
+        # Variables
+        self.photodiode_on = False
+        self.led_on = False
+        self.led = self.LED_OFF
+        self.stage = self.STAGE_OFFLINE
+        self.data = [[] for _ in range(3)]
```
FelicityMatt: magic number?? makes me go :grimacing:

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
         except Exception as e:
             self.get_logger().error(f"Failed to send led command over CAN: {e}")
-            response.success = False
 
-        return response
+
+        self.photodiode_light_blank_timer.cancel()
+        self.stage = self.STAGE_CALIBRATION
+
+        self.average_data[0] = fmean(self.data[0])
```
FelicityMatt: I have some concerns about averaging and standard deviations of data, there could be a lot of data points and will this be too time consuming? also is it something we even need? I thought science only wanted averages and std between readings at different spots within the sample. It's not something we had last design cycle, but we can do some testing and see how it goes

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
+        self.get_logger().debug(f"Entering online mode\nSending calibration mode average to gui: {self.average_data[1]}")
+
+        self.publish_msg(self.average_data[1], self.standard_deviation[1], True, True)
+
+        self.photodiode_online_timer.reset()
+
+
+    def photodiode_online_callback(self):
+        """
+        One-shot callback that deactivates photodiode after period is over
```
FelicityMatt: Do we need to send a CAN message to turn off the LED here?

src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -27,103 +36,298 @@
...
+        self.led = request.led
+
+        self.get_logger().debug(f"Activating photodiode {request.led}")
+
+        self.publish_msg(-1, -1, False, True) # -1 to not consider this value
+
+        return response
+
+
+    def photodiode_light_blank_callback(self):
```
FelicityMatt: is the LED ever turned on?

src/ros/rover/nova_interfaces/msg/NIRProbeData.msg
FelicityMatt: Still a bit confused honestly, would a new enum/field called `stage` or something be better to describe the stage instead of using `led_on` and `photodiode_on` which I find a bit confusing

FelicityMatt (commented):
Looking good, just some comments to make it even better and clearer. One other thing that we might want to clarify with science is do we want a constant stream of readings to be shown on the GUI or are they ok if the only readings they see are as a result of this service?

FelicityMatt:
moved into #271

PR #167, opened 2025-01-10
Packaged groot2

vic-bart:
## Description

- Groot2 is a BT visual editor for Nav2 and ROS2. The previous iteration, Groot, was packaged by Josh and used last design cycle, however it's no longer being maintained nor up-to-date with the BTs we're developing, which is why Groot2 needed to be packaged as well

## Steps to Test

- `nixfiles`, `nix-build -A pkgs.groot2`, `result/bin/groot2`, and it should run Groot2

## Memes

image


vic-bart:
tested by @TJ-Lee-M, works :+1:

PR #187, opened 2025-01-23
Adding temporary cameras-legacy alias for field tests

novarover:


jaredlandau:
tested in workshop, works

PR #190, opened 2025-01-26
Add generic store reset setting

FelicityMatt:
## Description

Added a tab within settings that allows generic stores to be reset. Most useful while developing features in the gui and working with generic stores. Could also be useful if changes have been made and localstorage has not been updated correctly.

To reset the user has to enter the correct name of the store, I did this instead of a dropdown/option picker to ensure you can't accidentally wipe data, you have to know what you are doing to use this feature.

## Changelogs

- Update Settings Model to include tabs
- Add a new tab for generic store resets
- Update docs to include this information

## Screenshots

Screenshot From 2025-01-26 17-12-12
Screenshot From 2025-01-26 17-12-16

## Steps to Test

Run the gui and click the top right settings button, also check localstorage and use the button to ensure it works.

## Memes

image

### DIFF
--- a/src/ros/nova-gui/docs/generic_store.md
+++ b/src/ros/nova-gui/docs/generic_store.md
@@ -12,6 +12,10 @@ A generic store is intended for use in situations where:
 For those familiar with `useState`, a generic store is meant to emulate that behaviour but
 with global, persistence and tab-sync properties.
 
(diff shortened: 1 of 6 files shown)


BaileyChessum (approved):
lgtm

PR #192, opened 2025-01-27
Creating a 'zero-arm' alias

jaredlandau:
## Description

yes...... another alias branch 🙏

## Steps to Test

zero the arm and then type 'zero-arm'
then type 'launch-arm' and test with joysticks

## Memes

image


jaredlandau:
Tested in the workshop, all functionality working as intended with no issues.

PR #193, opened 2025-01-28
Updating depthai versions

vic-bart:
## Description

:speaking_head: World, the time has come to... UPDATE DEPTHAI
- Depthai ROS doesn't archive previous versions of their packages. So when a new version is released, the old one is no longer retrievable. Yes, we do have a compiled version stored in Hydra, but the newer version is more likely to be ABI-compatible with the dependencies in Nixpkgs, so we should update regardless

## Steps to Test

- See here for instructions on how to make the changes
- To test, test the auto camera by running `ros2 launch auto_bringup camera.launch.py`

## Memes

Depthai ROS on their way to update the package:

### DIFF
--- a/nixfiles/packages/ros/depthai/default.nix
+++ b/nixfiles/packages/ros/depthai/default.nix
@@ -10,7 +10,7 @@
 
 buildRosPackage rec {
   pname = "depthai";
-  version = "2.28.0";
+  version = "2.29.0";
   buildType = "prebuilt";
 
   src = fetchurl {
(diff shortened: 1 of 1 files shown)


nixfiles/packages/ros/depthai/default.nix
```diff
@@ -23,12 +23,12 @@ buildRosPackage rec {
...
-      url = "http://packages.ros.org/ros2/ubuntu/pool/main/r/ros-jazzy-depthai/ros-jazzy-depthai_${version}-1noble.20240822.150456_amd64.deb";
-      hash = "sha256-4fAEmPSQJXQFl4efPTEt6C2C0/eoHy2tE6Z8CXNsYp4=";
+      url = "http://packages.ros.org/ros2/ubuntu/pool/main/r/ros-jazzy-depthai/ros-jazzy-depthai_${version}-1noble.20250102.150449_amd64.deb";
+      hash = "sha256-vUybJBr3QPwqZo7lvkbgV+0PwQNezbmJrxOfKSM1wqk=";
     };
     aarch64-linux = {
-      url = "http://packages.ros.org/ros2/ubuntu/pool/main/r/ros-jazzy-depthai/ros-jazzy-depthai_${version}-1noble.20240822.150457_arm64.deb";
-      hash = "sha256-qvAXejjxyVtimsg9Xx0fBQaJWsUtIzm7fgEzCDwSdcs=";
+      url = "http://packages.ros.org/ros2/ubuntu/pool/main/r/ros-jazzy-depthai/ros-jazzy-depthai_${version}-1noble.20250102.133606_arm64.deb";
+      hash = "";
```
hacker1024: Gotta get the Aarch64 hash too

hacker1024 (changes requested):

PR #200, opened 2025-01-31
Automatically launch new nova-gui service on startup

jaredlandau:
## Description

this may or may not work

## Steps to Test

need to rebuild and see if nova-gui service has been enabled and autostarted

## Memes

image


jaredlandau:
After speaking with Josh L, it has become apparent that we do not want to run a GUI service on startup for various reasons. Closing this PR and instead work will be moved to a new branch where a single command can call a script to launch GUI.

PR #209, opened 2025-02-11
refactor: removed breakpointHook from controller-manager

RohitPilakkat:
## Description

Removing breakpointHook from controller manager because it isn't needed


BaileyChessum (approved):
yep

vic-bart:
lgtm

PR #213, opened 2025-02-17
feat/bt-blackboard Variable Publisher

tarikthomas:
## Description

This PR introduces a Blackboard Variable Publisher node that extracts and publishes values from the behavior tree's blackboard, converting known variable types into human-readable formats. This allows for improved debugging, monitoring, and visualization of key behavior tree parameters.

## ROS

- publishes a ros topic called blackboard_debug

## Steps to Test

currently the node is added to search_for_ar_tags.xml tree. To test if it's working run auto_bringup everything.launch.py and in rviz make sure you run nav through poses. Then check if the ros topic 'blackboard_debug' is being published and make sure it is echoing the correct data by checking ros2 topic echo /blackboard_debug. Also when echoing not all the data shows up so to see everything use ros2 topic echo /blackboard_debug --full-length

This node is currently only on behavior_tree/other/search_for_ar_tags.xml however this tree should launch by defualt.


vic-bart:
Code rebased, tested, and reviewed, looks good. There are some changes I would make regarding implementation, but that would require this code to be used with the rest of the stack so we can determine what blackboard variables we want to see through practice, which is only possible after being merged.

PR #217, opened 2025-02-18
fix: temporary fixes to make arm show on Banksia

RohitPilakkat:
## Description

The move from Waratah to Rover7 (Banksia) caused this part of the urdf file to break.
It's still a little messy as of now since the new rover comes with a payload base attached to it but this can be fixed later.

## Screenshots

Screenshot From 2025-02-17 22-44-02

## Steps to Test

`ros2 launch nova_bringup urdf.launch.py arm:=true`


BaileyChessum (approved):
ok

PR #224, opened 2025-02-22
Add GUI aliases

FelicityMatt:
## Description

I participated in a comp sim this morning where I helped set up the gui for base station and realised how hard typing out all the long commands can be so decided to add these aliases to help speed up the process and reduce time wasted with the GUI setup.

## Changelogs

- added two aliases `gui-shell` and `gui-link`.
- updated the nova-gui readme and `rover-help` with information about these aliases.

## Screenshots

image

## Steps to Test

Follow the new rover help section

## Memes

image


jaredlandau:
LGTM, this is a great idea

BaileyChessum (approved):
Would be cool to have an alias that does all three in one. But, looks good for now!

PR #226, opened 2025-02-24
Let science run their payload at the beach (hopefully)

jaredlandau:


jaredlandau:
Science ran successfully at the beach, resolved. Long-term fix being implemented at fix/science-launch.

PR #233, opened 2025-02-27
Update science launch

FelicityMatt:
## Description

Updated the Auger and Science Platform ROS2 nodes to accept CAN IDs as parameters so that both the ARC and URC launch files can pass in different CAN IDs for the different payloads.

- The URC launch file has been tested with the physical system and works.
- The ARC launch command works locally with no issues (and worked with `urc_auger` and `urc_analysis_arm` with hard coded CAN IDs).

Also includes a fix to the `PumpsWidget` which caused the URC Science Dashboard page to fail to load.

Merging this means that the branches `fix/science-beach` and `feat/arc-science-testing` can be removed as this PR covers those changes.

## Changelogs

- parameterise [urc_analysis_arm.py](https://github.com/MonashNovaRover/nova/compare/master...fix/science-launch#diff-ff22be7824739c6fa783d073f39a1cb4a92a8b2ffef63dc37b49fc48e9fc2ef9) and [urc_auger.py](https://github.com/MonashNovaRover/nova/compare/master...fix/science-launch#diff-80606d37e029a640e95c567b87f27a49e94851a50abd825a6bad042258caff0b).
- ARC launch file now uses `urc_analysis_arm.py` and `urc_auger.py`
- Added launch arguments to both ARC and URC launch files.
- Removed old ARC `auger.py` and `analysis_platform.py` which are no longer used.
- Updated some other URC CAN IDs.

## Steps to Test

- ws-build
- run 
  - `ros2 launch nova_bringup arc_science.launch.py`
  - `ros2 launch nova_bringup urc_science.launch.py`


FelicityMatt:
The GUI builds locally for me, I'll try the workflow again in a bit.

BaileyChessum (commented):
looks good :)

PR #234, opened 2025-02-27
feat: Added urdf-modifier to urdf-tool

V01DBREAKER:
## Description

Adds urdf-modifier to the urdf-tool shell alias

the urdf-modifier is a command line script that takes any urdf/xacro/xml file and will:
- Recalculate all link inertia tags which have an attached stl model
    - To address incorrect inertias by onshape-to-robot
- Remove _continuous from any joint tag name
    - To address onshape-to-robot requiring _continuous being suffixed to any continuous joint you wish to make
- Rewrite any mesh urls from package:// to file://
    - Since this caused issues with gazebo
 
 This tool should streamline the export of future URDFs

Inertias have been tested and are comparable to https://www.hamzamerzic.info/mesh_cleaner/
Which is the tool we used to fix the arm-urdf inertias.

https://github.com/V01DBREAKER/urdf-modifier - Source code can be found here

## Usage:
`urdf-tool`
`urdf-modifier <urdf-file> [mesh-directory]`

## Memes

image

### DIFF
--- a/nixfiles/modules/home/macros/urdf-tool.nix
+++ b/nixfiles/modules/home/macros/urdf-tool.nix
@@ -51,11 +51,34 @@ let
       SECRET_KEY = "";
     };  # If the file does not exist, use set with empty values to avoid errors
 
+  urdf-modifier = pkgs.python3Packages.buildPythonApplication rec {
+    pname = "urdf-modifier";
+    version = "1.0";
+
+    src = pkgs.fetchFromGitHub {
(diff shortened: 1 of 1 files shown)


vic-bart:
Tested and works, lgtm

RohitPilakkat:
I've used this commit in a few other branches and it seems to work fine. Merging :)

RohitPilakkat (approved):

PR #237, opened 2025-02-28
Update AP Code

### LINKED ISSUE #223: Update AP Code
AP ros2 nodes currently use an old version of the python control classes since they were changed between ARC and URC last design cycle.

AP code now needs to be updated to use the new control classes, this includes:
- Scraper
- Tile Placer

A currently working version exists on the branch [`/fix/scraper-x-ids`](https://github.com/MonashNovaRover/nova/compare/master...fix/scraper-x-ids)

FelicityMatt:
## Description

 
Updated scraper and tile placer python control classes to be up to date with the current python control code.

Tested at a field test today and it worked!

Fixes: #223 

## Changelogs

- Updated `scraper.py`
- Updated `tile_placer.py`

## Screenshots

See Slack message

## Memes

image

### DIFF
--- a/src/ros/rover/excavation_construction/excavation_construction/tile_placer.py
+++ b/src/ros/rover/excavation_construction/excavation_construction/tile_placer.py
@@ -15,25 +15,25 @@ PACKAGE: control
 COMMAND: ros2 run control tile_placer.py
 RUN ON: Rover
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-AUTHOR(S):	Tristan Clark
+AUTHOR(S):	Tristan Clark, Felicity Matthews
 CREATION:	02/02/2024
-EDITED:		09/03/2024
+EDITED:		20/02/2025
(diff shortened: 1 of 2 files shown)


src/ros/rover/excavation_construction/excavation_construction/tile_placer.py
```diff
@@ -157,40 +117,21 @@ def update_tile_placer(self, joystick_r: InputJoystick):
...
-        self.get_logger().debug("called l")
-
-        joystick_l = msg
-
-        self.update_joystick_lock(joystick_l)
-
-        if self.check_joystick_lock():
-            return      
-            
         self.velocity_multiplier = abs(joystick_l.ax_slider)
```
BaileyChessum, line 125: does `self.update_tile_placer(joystick_r)` need to be called here too?

BaileyChessum (approved):
If tested and work good, looks good

src/ros/rover/excavation_construction/excavation_construction/tile_placer.py
FelicityMatt: for tile_placer no I don't think so since the only buttons controlling the forklift are on the right joystick. The left joystick only changes the velocity by the looks of it which seems correct to me.

PR #276, opened 2025-03-08
[WIP] Updates to latest packages before hydra vomitting

rotaryViper:
The hydra vomitter server doesn't always build the most up to date packages. This request makes the pc do a git pull first and then checkout master, then nix-build.


rotaryViper:
To do:
Write a script that will run hydravomit if git pull has changed.

PR #285, opened 2025-03-09
New camera and better GUI aliases

FelicityMatt:
## Description

I created new camera aliases that should work better (?) (or at least the same as) the `cameras-legacy` alias that refer to the up to date cameras2 build.

Updated the rover help page with the new information.

### Three new commands:
- `launch-cameras` should be succeeded with a payload:={payload} parameter
  - eg `launch-cameras payload:=ec`
  - ** note this command can only be used on a jetson as the parameters will be different on another device
  - This is the command that was used for the E&C payload camera testing.
- `launch-cameras-all' same as above but will auto start all cameras (can be a bit iffy)
- `gui-run` will launch the gui from anywhere

### Updated command:
- `gui-link` this command can now be called anywhere.

** note that `yarn install` still needs to be called in the right directory

## Screenshots

image

## Steps to Test

- `sudo nixos-rebuild switch` and try them yourselves.

## Memes

image

### DIFF
--- a/nixfiles/doc/rover-help.md
+++ b/nixfiles/doc/rover-help.md
@@ -38,10 +38,9 @@ Launch the following on the metabox.
 In one terminal:
 1. Try the 'gui-shell' alias otherwise:
      nova-shell -A pkgs.ros.nova-gui
-3. Try the 'gui' alias otherwise:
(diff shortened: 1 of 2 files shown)


jaredlandau:
LGTM, these are awesome changes!

PR #296, opened 2025-03-12
Fix/arm resolver refactor

### LINKED ISSUE #298: Arm Resolver Refactor
The way the arm's control code handles resolver output involves manual polling. This causes a lot of overhead and delays. The code also does not handle messages out of order well and just discards these values.

Refactor the code to handle automatic data streams and improve how resolver values are captured.

RohitPilakkat:
## Description

resolver publisher now sets resolver output to automatic mode and does not need to manually poll before receiving a response anymore.

According to the notion documentation on CAN Bus standards, 0x0A2 sets the resolvers to automatic mode. This allows the resolvers to output a constant stream (Currently at 50Hz) of data that we now capture and assign to the correct joint elements instead of having to manually poll the resolvers.

## Steps to Test

1. Run the arm control stack.
2. Observe CAN messages along can1 using `candump can1`
3. Resolvers will output constant feedback at 50Hz
4. Arm should be controllable using standard controls.

### DIFF
--- a/src/ros/rover/arm/arm/arm/resolver_publisher.py
+++ b/src/ros/rover/arm/arm/arm/resolver_publisher.py
@@ -127,11 +127,6 @@ class ResolverTransceiver(CANTransceiver):
         if not self.transmit(enable_message):
             self.logger.error("Failed to send auto mode enable command!")
 
-    def enable_auto_timer_callback(self):
-        """Re-send the 0x0A2 command every 1 second."""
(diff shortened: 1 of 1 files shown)


RohitPilakkat:
Rviz display probably needs some fixing

BaileyChessum:
is this ready for review? we should aim to get this in ASAP

RohitPilakkat:
This is good to merge since it has been tested several times during comp sims.

PR #299, opened 2025-03-12
Auto mount URDF!

V01DBREAKER:
## Description

Adds the auto mount to the URDF
Makes the base rover use just the main body without the arm mount
Adds the adjustable angle to the auto mount 

## Screenshots
image

## Steps to Test

Could someone test if this affects the auto stack at all?
To launch just the urdf to take a look: `ros2 launch auto_bringup gazebo.launch.py`

## Memes
image


vic-bart:
Need to merge this in feat/bt before master

PR #301, opened 2025-03-13
Feat/urdf fixes

vic-bart:
## Description

Adds the auto mount to the URDF
Makes the base rover use just the main body without the arm mount
Adds the adjustable angle to the auto mount 

## Screenshots
image

## Steps to Test

Could someone test if this affects the auto stack at all?
To launch just the urdf to take a look: `ros2 launch auto_bringup gazebo.launch.py`

## Memes
image


V01DBREAKER:
image
check this before merging

Issue #248, opened 2025-03-04
Reolink Camera Support


Redecorating:
stream: 
`ffplay -rtsp_transport tcp -i rtsp://admin:$PASSWORD@10.0.1.100:554/Preview_01_main -flags low_delay -fflags nobuffer -framedrop`

a silly way to get the password without putting it in git history could be: `sudo cat "/etc/NetworkManager/system-connections/Nova Mission Control Centre.nmconnection" | grep psk= | cut -d= -f2`

Don't know how hard it would be to make ffmpeg output to the gui - but this may not be needed

Controlling camera tilt/pan: maybe use something like this https://github.com/quatanium/python-onvif/blob/master/examples/continuous_move.py

  
  Tags for Notion Integration:

Redecorating:
should make it controlled through ros https://robotwebtools.github.io/rclnodejs/docs/0.32.0/Service.html

PR #302, opened 2025-03-13
Feat/reolink

Redecorating:
## Description

nodejs script to control reolink camera
also updated some of the aliases for opening the camera feed -- need to make these a script so we can support upside down

#248 

## Steps to Test
plug in camera, there is a 12v power supply that works for it, i put a sticker saying "works for reolink" on this. Plug into ethernet as well, set your static ip to smth like 10.0.1.1/24.

```bash
cd /home/nova/nova/nixfiles/packages/other/reolink/src
nix-shell -p nodejs_23
npm install .
node ./main.js
```

im somewhat at a loss on how to integrate this with nix so it gets installed. You could run this on the rover so that the laptop isn't sending commands itself during auto task.

also for video playback:
use `nix-shell -p ffmpeg`, then `reolink-high`. I'm not sure how to make sure ffmpeg is installed for all the computers.

video: https://monashnovarover.slack.com/archives/C088X11KQ5R/p1741859902144639?thread_ts=1741836236.742639&cid=C088X11KQ5R

## Memes
image


Redecorating:
made some silly aliases to get around me not being able to build these with the nova stuff, which hopefully is good enough to use for the fieldtest

you can use `reolink-ctl` to use your keyboard to control the camera
    and `reolink`/`reolink low`/`reolink low 180` etc to open the camera
    feed with (default) high resolution, or alternatively low res, and
    optionally with 180 degree rotation. order of arguments doesn't matter.

rotaryViper:
May I ask why in the js file,
line 131 is STOP_DELAY_MS/2
and line 144 is 2*AUTO_TIMING
surely there is some spaghetti there 🫠

Redecorating:
> May I ask why in the js file,
> 
> line 131 is STOP_DELAY_MS/2
> 
> and line 144 is 2*AUTO_TIMING
> 
> surely there is some spaghetti there 🫠

after not receiving move commands for STOP_DELAY_MS, the camera stops moving. For auto mode it sends commands at twice this rate (half the period) so it won't stop. The 2* in the other line balances that out so the whole panning to one side phase has the correct duration (I forgot which way around the operands for division are when i wrote this so my math might be wrong)

Redecorating:
the js code is somewhat messy and convoluted because it is a rushed job adapting an example script instead of writing code from scratch that suits our usecase really well

BaileyChessum (approved):
looks good

PR #319, opened 2025-03-16
Spinny Part Positional Control

FelicityMatt:
## Description

Sprinny part now uses positional control instead of continours


BaileyChessum (approved):
Looks fine to me? We will test on rover

PR #320, opened 2025-03-16
Update python node to take multiple readings

FelicityMatt:
## Description

- NIR probe now takes multiple readings and returns the average and standard deviation.

### DIFF
--- a/src/ros/rover/science/science/nir_probe_publisher.py
+++ b/src/ros/rover/science/science/nir_probe_publisher.py
@@ -203,7 +203,16 @@ class NIRProbePublisher(Node):
             return
 
         if self.active_photodiode == self.PHOTODIODE_OFF:
-            self.get_logger().warn(f"Reveived frame when active_photodiode is off {frame}")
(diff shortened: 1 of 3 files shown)


src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -108,12 +122,71 @@ def publish_msg(self):
...
+            self.last_stdv = 0
+        else:
+            self.last_average = 0
+            self.last_stdv = 0
+
+        # reset variables
+        self.readings = []
+        self.active_photodiode = self.PHOTODIODE_OFF
+
+        self.publish_msg()
```
BaileyChessum, line 187: for future, we should check that we are using the same LED when accumulating values

BaileyChessum (approved):
I don't notice any issues from a cursory glance

BaileyChessum (approved):
update the gui :)

FelicityMatt:
gui builds :)

Issue #289, opened 2025-03-10
Add twitching controls to the Tile Placer


BaileyChessum:
Currently, to control the tile placer, we hold a button, and the tile placer moves up/down until we let go of that button. This has problems, because if there is lag, it will continue to just keep going up and down until we get connection, and messed up our alignment multiple times in prior comp sims.

Similar to science analysis platform control, we want to be able to 'twitch' from the ROS2 node. This is just turning on the motor, waiting a fixed amount of time, and then turning off the motor. So, in effect, pressing the button moves a predictable and fixed distance.

  
  Tags for Notion Integration:
  tag:software;tag:rover

rotaryViper:
Where the file at?

BaileyChessum:
https://github.com/MonashNovaRover/nova/blob/master/src/ros/rover/excavation_construction/excavation_construction/tile_placer.py

rotaryViper:
https://github.com/MonashNovaRover/nova/pull/321

PR #321, opened 2025-03-16
[NEEDS TESTING] Feat/tile placer twitch

rotaryViper:
It should twitch

Tags for Notion Integration:
tag:science;tag:qol


rotaryViper:
https://github.com/MonashNovaRover/nova/issues/289

FelicityMatt:
@rotaryViper what's the current status on this?

rotaryViper:
It just has to be tested with tile placer but it should work. It was for ARC but I didn't know how to test it.

FelicityMatt:
We will move to python control2 so this wil be redone in future and is not worth testing.

Issue #310, opened 2025-03-14
Scraper left joystick fwd/back is flipped


BaileyChessum:
we need to flip forward and backwards on the scarper needs to be inverted for scraper.py.

rotaryViper:
https://github.com/MonashNovaRover/nova/pull/322

PR #322, opened 2025-03-16
Flipped forward and backward on left joystick

rotaryViper:
Left joystick is flipped

### DIFF
--- a/src/ros/rover/excavation_construction/excavation_construction/scraper.py
+++ b/src/ros/rover/excavation_construction/excavation_construction/scraper.py
@@ -243,7 +243,7 @@ class ScraperNode(JoystickControllerNode):
             Left Joystick stick x-axis = direction / velocity
         """
         self.arm.update_velocity(abs(joystick_l.ax_stick_x) * self.velocity_multiplier)
-        self.arm.update_direction(self.ARM_FORWARDS if joystick_l.ax_stick_x >= 0 else self.ARM_BACKWARDS)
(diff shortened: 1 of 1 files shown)


rotaryViper:
https://github.com/MonashNovaRover/nova/issues/310

src/ros/rover/excavation_construction/excavation_construction/scraper.py
```diff
@@ -243,7 +243,7 @@ def update_arm(self, joystick_l: InputJoystick):
             Left Joystick stick x-axis = direction / velocity
         """
         self.arm.update_velocity(abs(joystick_l.ax_stick_x) * self.velocity_multiplier)
-        self.arm.update_direction(self.ARM_FORWARDS if joystick_l.ax_stick_x >= 0 else self.ARM_BACKWARDS)
```
FelicityMatt: if this logic is correct you can also reverse the directions of these variables defined here: https://github.com/MonashNovaRover/nova/blob/424de9c0fd8a7aa98f45c0cedbdb3cb91bfa42e5/src/ros/rover/excavation_construction/excavation_construction/scraper.py#L58-L59
It's probably best for this logic to remain accurate (ie if input is this move forward else back) and switch the direction if that makes sense, should do the exact some thing

FelicityMatt:
also while we're here could you change the default starting velocity to something other than 0, maybe like 0.3 or something, cause i think we keep thinking the code is broken because the velocity is set to 0 at the start

rotaryViper:
Very logical, will make the logic better

FelicityMatt (approved):

PR #324, opened 2025-03-17
Update microscope view

FelicityMatt:
## Description

Tidied up the microscope view for use with ML (while keeping thresholding functionality)

ML output field is just an input field where we can copy paste ML terminal output, just somewhere to store the result thats in the GUI, could be considered the same as an excel spreadsheet.

Can store ML output for each site.

## Changelogs

- moved existing threshold to a tab.
- added a tab that contains a microscope camera component.
- added an input for somewhere to put ML output.

## Screenshots

Screenshot From 2025-03-20 17-13-51
Screenshot From 2025-03-20 17-13-43


TJ-Lee-M:
looks good! But not sure what purpose of IllmeniteMLResult is

BaileyChessum:
How is the ML thing used?

FelicityMatt:
It's just a field on the gui where we can copy paste output from the ML model, it's stored in generic store so will persist and also looks a bit nicer and uses the space that we have if we're not using thresholding

TJ-Lee-M:
Looks good! :saluting_face:

BaileyChessum (approved):
sure, looks good.

PR #326, opened 2025-03-17
Filter out non-raw cameras in the directory server

hacker1024:
## Description

The Arducams have two UVC devices that can be opened simultaneously. The camera directory server expects one UVC device per USB device, though. Luckily, we only need one of them, so filter the one we don't need out.

A better long term fix will be to update the directory server API to support multiple UVCs per camera.

## Memes

image


BaileyChessum (approved):
looks good! we should raise an issue to do the suggested future changes

V01DBREAKER:
This PR caused an issue with building cameras2legacy and has been reverted. Maybe think about adding this back in future?
image

V01DBREAKER:
#526

PR #327, opened 2025-03-17
Send CAN commands only when frame changes

FelicityMatt:
## Description

- added a toggle to the controller class that can switch between sending can commands continuously or only when the frame changes.


TJ-Lee-M:
lgtm, probably worth waiting to merge until testing on rover tmr?

FelicityMatt:
just tested locally with scimbal cam and it works, only sending can commands when something the frame changes.
image

BaileyChessum (approved):
looks good!

PR #331, opened 2025-03-18
Add another terminal with Shift+Ctrl+F support

BaileyChessum:
## Description

This adds Ptyxis, another terminal

Pros:
- Shift+Ctrl+F works

Cons:
- Has a bunch of padding for some reason, which I don't like. Oh well

## Screenshots

image


Nova-Jetson (approved):
I am accepting my own PR, thanks

PR #334, opened 2025-03-19
Ilmenite MVP

jaredlandau:
## Description

Will work with CUDA when feat/metabox-cuda-support is merged. As a backup, predict_ilmenite_cpu.py has been created to force CPU usage instead if CUDA doesn't work for whatever reason. This is MVP for ilmenite, further changes coming shortly though, which will extend functionality and make it easier to use.

Bailey has created a shell with all Python dependencies (including pytorch) to make this script easy to use. Everybody type 'thank you Bailey' for good luck!


TJ-Lee-M:
Lgtm! :)

FelicityMatt (commented):
lgtm

jaredlandau:
Tested in workshop and works, merging so that we have MVP for final science comp sim tomorrow. Will open a future PR to add support for running multiple models at once, and improve existing model performance.

PR #335, opened 2025-03-19
more reliable can, added support for can2

jaredlandau:
## Description

* can0 starts automatically in drive.launch.py
* can1 starts automatically in arm.launch.py and arc_science.launch.py
* can2 support added to can.sh

will test this asap (they took the metabox away from me to test kiln)


BaileyChessum (approved):
yep

PR #336, opened 2025-03-20
Science Cleanup

FelicityMatt:
## Description

Small science changes to tidy things and add logging and update mappings.

## Changelogs

- sweeper controls changed to right buttons to avoid button clash
- logging and parameters added to spinny part
- camera science arc payload changed from science_arc to science-arc to remove confusion (now same as `launch-science-arc`)
- removed microscope from launch file
- undo drill direction change since it changes through buttons.
- update kiln max temp to be inline with electrical max temp.


FelicityMatt:
Will test tomorrow to confirm everything works.

BaileyChessum (approved):
looks fine

PR #340, opened 2025-03-20
Default configuration for camera view mast forward and arm periscope

RohitPilakkat:
## Description

I have rotated camera view "mast forward" by 180 and flipped camera view "arm periscope" as per Paola and Jared's preferences for the post landing task.

## Steps to Test

try gui and see if the default camera views are correct (mast forward should be right side up and arm periscope should already be set to flipped)


FelicityMatt (approved):
lgtm

PR #344, opened 2025-03-22
invert w/s for scimbal wasd

TJ-Lee-M:
## Description

fixed wasd so w is up, s is down

## Changelogs

Changed the sign of the increment for the w and s cases.

## Memes
Scimbal when you tell it W: [goes down]
:(


TJ-Lee-M:
It might've made make more sense to update scimbal.py tilt definition, so that the generic gimbal camera componant isnt hardcoded to the scimbal's directions (in case we ever wanted to reuse gimbal component)?? But not relevent in current scope and this is an easy fix for ARCh

FelicityMatt (approved):

FelicityMatt:
I think this is fine for now, the gimbal cam is upside down atm and depending on what orientation it is positioned will determine which way it looks so I think its fine to be on the gui

PR #346, opened 2025-03-22
Feat/auto reolink

Redecorating:
## Description

- Tried to reduce motion sickness (slow down auto mode pan)
- added support for saving video feed to disk

Please let me know if:
- When you do `reolink save`, have that run for 30-40 minutes as the rover does its task, then press q/ctrl-c, does it take a long time to exit? (which would indicate that it is not fast enough at saving the video)
- My attempts to slow down the panning just give you more motion sickness
- it would be worthwhile to make auto mode do tilting or smth

## Steps to Test

tldr:
to save recording, run on rover: `reolink save`, press `q` on keyboard when done, it will print where it saved the file.

to adjust auto mode speed params, press 1234 and hope for the best

saving to file:
```
    reolink: allow saving video to disk instead
    
    Mainly for auto to run on rover so that they can look at good video
    later and don't need to stream it while the auto stack is using most of
    the bandwidth.
    
    Use `reolink save`. The default filename will be
    ~/recordings/reolink_$CURRENTTIME.mp4, but you can specify a filename
    with `-o filename.mp4`.
    
    Once you copy the file from the rover to your laptop, you can play it with
    `ffplay /path/to/file.mp4` (or use your prefered video player program).
    
    Also you don't _need_ to use mp4, other formats will probably work too.
    
    Also added a --uri option for testing with other uris (eg for testing at
    home if you have an IP camera)
```

in auto task i guess you would use scp or smth after the recording is done to move the video to your laptop and then open it.

make auto mode not give you motion sickeness hopefully:

```
    reolink-ctl: try to add speed control for auto mode
    
    its not very good
    
    we control speed by sending a new move command when the camera is moving
    this makes the camera stop then start moving. If you do this frequently,
    it is really slow, if you do it infrequently it is normal speed but it
    moves in bursts.
    
    Use keys 1234 to modify the two parameters for this.
    
    if you want the previous mode, use speed=6, pan duration=6 i think
```

## Memes

the new "remove french language pack" command: `ffplay "rtsp://rtspstream:6m6Ys4UIDvErTYlwoGxBk@zephyr.rtsp.stream/pattern"` (this kills my home router for some reason :/)


Redecorating:
if someone merges, probably don't squash because the first commit might need reverting

Redecorating:
i think this has been squash merged into feat/bt already. if there are merge conflicts then take the version from feat/bt

PR #347, opened 2025-03-22
Move camera tabs

FelicityMatt:
## Description

When cameras are running we found that we had to scroll down to see drive telemetry which can be a bit stressful in the middle of a comp sim where every moment counts. I wanted a way to reduce the size of the top bars and bring the components up and thought that merging the camera tabs and buttons was a great way to do this.

## Changelogs

- converted camera tabs to use the segmented picker

## Screenshots

Before:
Screenshot From 2025-03-23 10-35-00

After:
image

## Memes

image


jaredlandau:
lgtm!

BaileyChessum (approved):
Seems good, but we should check with real cameras to ensure state isn't randomly reset

FelicityMatt:
tested with the camera stack on my laptop and it worked better than the original version

PR #348, opened 2025-03-23
added ffmpeg and libreoffice

RohitPilakkat:
## Description

libreoffice and ffmpeg 

I added it to home/workspace so it only builds when the nova workspace config is enabled.


BaileyChessum (approved):
well give it a try

PR #349, opened 2025-03-23
Feat : VIO Approach

KABILAN235:
## Description

Switched from F2F (Frame to Frame) to OpenVINS which is a Visual Inertial Odom Approach. The IMU Transformer for some reason is cooked and this is the right opportunity to switch to VIO and remove IMU from EKF

## Memes

Photo from 2025-03-23 16-33-56 967408


KABILAN235:
For this to work,  VINS-Fusion has to be compiled and added to build dependencies of rtabmap. I'm closing this PR for now because there's no time to compile VINS-Fusion

PR #350, opened 2025-03-23
Autosave NIR readings

BaileyChessum:


FelicityMatt (commented):
lgtm and we tested it right?

novarover (approved):
tested with the nir probe and it works! very helpful, I've also just removed the disable option from the button so that if the nir board doesn't respond we can try again, this has worked so far in testing

PR #356, opened 2025-03-26
add target temp to generic store

TJ-Lee-M:
## Description
added target temp to generic store

## Changelogs
Created targetTemp generic store with default value (150) in RootReducer.ts.
Defined type in RootState.ts.
Updated kiln widget to use generic store.

## Steps to Test
change kiln target temp with slider on microscope gui page and refresh to see it is retained. Repeat with typing value input instead of click/dragging slider


FelicityMatt (approved):
lgtm

PR #359, opened 2025-03-27
Feat/bt load waypoints on failure

novarover:


novarover:
lgtm, written/tested by @tytia and @tarikthomas - Victor

PR #372, opened 2025-04-05
Deleting feat/bt

vic-bart:
## Description

PR to merge in autonomous development that has mainly occurred on `feat/bt` into `master` post ARCh 2025.


vic-bart:
Tested in sim, still seems to work. I'd like to test IRL and with the reolink as they changed significantly on the rebase, but I think the rebase was correct; the auto mount is disassembled; I want this PR done ASAP.

vic-bart:
Read through all the changes, couldn't tick off files viewed due to the sheer magnitude of this PR, but lgtm :+1:

PR #373, opened 2025-04-06
Make the IK controller chainable with a distinct Twistmapper controller

### LINKED ISSUE #363: Split the Twistmapper from the IK Controller, to become a controller chain
Currently, the twistmapper for arm is a part of the nova_ik_controller. We need to 
- Move logic for the twistmapper into its own controller
- Make the nova_ik_controller a chainable controller that accepts the pose (relative to the IK origin) to move the arm to.
  - conscequently, the nova_ik_controller should have any dependency on tf2, (apart from maybe safety in the future, safety should currently be the responsibility of the twistmapper or whatever else provides the target pose)
  - this allows us to use a different pose source for auto typing
- Add the twistmapper as a controller in the controllers.yaml and to the controllers listed in teleop_arm_joy, so it gets spawned with the IK stuff as appropriate

BaileyChessum:
## Description

I have made the IK controller yet another chainable controller. This allows us to swap out the source of the input target pose.

## Changelogs

- Add nova_twistmapper, encapsulating the logic for the twistmapper previously in nova_ik_controller
- Make nova_ik_controller chainable
- Changed launch configs to include the new chain
- Fixed bugs in teleop_arm_joy with enabling/disabling controller chains

## Steps to Test

after `ws-build`ing

```sh
./result/bin/ros2 launch auto_bringup mock.launch.py arm:=true
```

```sh
./result/bin/ros2 run joy joy_node
```

```sh
./result/bin/ros2 run teleop_arm_joy teleop_arm_joy_node --ros-args --params-file /home/nova/nova/src/ros/rover/teleop_arm_joy/config/ps5.config.yaml
```

```sh
./result/bin/rviz2
```

In rviz, add a RobotModel with /robot_description as the topic. Press the unlock button, and move around

### DIFF
--- a/nixfiles/packages/ros/nova-workspace/default.nix
+++ b/nixfiles/packages/ros/nova-workspace/default.nix
@@ -42,6 +42,7 @@
 , nova-utils ? throw "nova-utils is needed, but not available!"
 , nova-arm-controller ? throw "nova-arm-controller is needed, but not available!"
 , nova-ik-controller ? throw "nova-ik-controller is needed, but not available!"
+, nova-twistmapper ? throw "nova-twistmapper is needed, but not available!"
 
   # Configuration options
   ## Include graphical applications in the workspace.
@@ -86,7 +87,8 @@
       nova-utils
       reolink
       nova-arm-controller
-      nova-ik-controller;
+      nova-ik-controller
+      nova-twistmapper;
   }
 
   ## Extra packages to add to the workspace.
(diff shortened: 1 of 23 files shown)


src/ros/rover/controllers/nova_ik_controller/package.xml
```diff
@@ -4,7 +4,7 @@
   <name>nova_ik_controller</name>
   <version>0.0.0</version>
   <description>TODO: Package description</description>
-  <maintainer email="arbab.fahm@gmail.com">nova</maintainer>
+  <maintainer email="novaroverteam@monash.edu">nova</maintainer>
```
BaileyChessum, line 7: we should add peoples names as header comments in each file, and just put nova here, as we will all eventually leave the team, and wont be maintaining this after

Redecorating (approved):
works on my computer

RohitPilakkat (approved):
Tried on my system and the controllers seem to be swapping correctly.

PR #379, opened 2025-04-08
Fix/old arm speed

RohitPilakkat:
## Description

This gives the arm finer controls for slow arm control for operators.

It was tested on the TX2s when we used those and seemed to work fine.
Has not been tested on the Orin yet.


Redecorating (approved):
This seems sane, LGTM.

PR #382, opened 2025-04-10
[OLD] Working NIR many readings code

FelicityMatt:
## Description

Working code to take many readings within the NIR Probe


src/ros/rover/science/science/nir_probe_publisher.py
```diff
@@ -45,8 +45,8 @@ class NIRProbePublisher(Node):
     NIR_PROBE_ID = 0x0F0
 
     NIR_RECEIVE_ID = [
-        PHOTODIODE1_ID := 0x4F0,
```
FelicityMatt, line 48: What is currently in master is correct

PR #410, opened 2025-04-11
rfid null bytes

RohitPilakkat:
allows messages to have null bytes interplaced and still reads the whole message


BaileyChessum (approved):
Looks great!

PR #418, opened 2025-04-12
Change default GUI route to URC cameras

FelicityMatt:
## Description

Updated the default route of the gui to navigate to URC camera page which should help reduce the number of clicks needed on launch.

## Memes

image


TJ-Lee-M (approved):
:+1:

PR #420, opened 2025-04-12
fix: Change rover7 to banksia.

V01DBREAKER:
Change gazebo rover model name from Rover7 to Banksia. 
Make nova_bringup urdf.launch.py use banksia_arm.
Change rover_description/rover7 to banksia and update filepaths

- [x] Check auto stack still runs with model rename in Gazebo. 
- [x] Check URDF has no issues 
- [x] Check nova_bringup has no other issues


RohitPilakkat (approved):
everything works for me too

PR #421, opened 2025-04-12
Feat/arm joint limits

### LINKED ISSUE #208: Add safety limits to nova_arm_controller
We need to add appropriate safety limits to the arm ros2control controller.

This can include angle limits, speed limits, position limits, auto halt commands and anything else required for general safety and usability of the arm for testing and operation.

### LINKED ISSUE #241: nova_arm_controller Position limits
In nova arm controller, ~~add state interfaces for position~~, and use it to implement limits on position for each joint.

### LINKED ISSUE #242: nova_arm_controller Velocity limits
In nova arm controller, implement limits on the output velocity. Prevent velocity values from being used that exceed limits. 

For now, we can specify limits by adding to the [controller parameters](https://github.com/MonashNovaRover/nova/blob/arm-master/src/ros/rover/controllers/nova_arm_controller/src/nova_arm_controller_parameter.yaml).

This could be done with the `speed_limiter.cpp`, which would be very nice as it could also apply limits to acceleration and jerk, helping to keep the arm more stable and precise.

We should try to allow this mechanism to apply when using both
- [ ] position control, and
- [ ] velocity control

For velocity control, this would be done using `speed_limiter.cpp` as normal. This should be done first, before trying with position control.

## Limiting velocity in position control mode

However, position control will be more complicated, as we are given target positions rather than velocities directly. 
This likely means 
1. Converting the input target position to a displacement (`Δx = x - x_previous`), alongside stored displacements from the last few updates
2. Deriving velocities from those displacements (`v = Δx / Δt`)

So, following the above rules, we calculate the values to give to `SpeedLimiter::limit` as:
      - `v  = (input target position - current state interface value) / Δt` 
      - `v0 = (current state interface value - previous update()'s state interface value) / Δt` 
      - `v1 = the previous update()'s value of v0` 

Then, we follow the steps backwards to get the limited target position from our newly limited `v`:

3. Convert velocity back into displacement (`Δx =  v * Δt`)
4. Convert the displacement into a position to find our velocity-limited target position (`x = x_previous + Δx`)
    - In this case, `x_previous` is the current position state interface value.

---

You can store all the necessary state for remembering things between updates inside of the `JointHandle` struct:
- the `SpeedLimiter`
- the `previous update()'s state interface value`
- the `previous update()'s value of v0`

### LINKED ISSUE #377: nova_arm_controller Position limits in velocity control mode
Implement velocity limits in nova_arm_controller when using velocity control (`params_.use_position_control` is false).

Effectively, we just need to read the current position, and ensure we don't send velocity commands that continue past that position.

### LINKED ISSUE #378: nova_arm_controller Position limits in position control mode
Implement velocity limits in nova_arm_controller when using position control (`params_.use_position_control` is true).

Effectively, we just need to ensure we don't send positions that exceed the limits. This should be the simplest of all the limits.

Redecorating:
## Description

add position/velocity/acceleration limits to nova_arm_controller by using SaturationJointLimiter. Jerk limits are not set up yet.

these can be configured in `src/ros/rover/controllers/nova_arm_controller/src/nova_arm_controller_parameter.yaml` or at runtime with stuff like:

```
./result/bin/ros2 param set /nova_arm_position_controller joint_limits.J1.max_velocity 0.1
./result/bin/ros2 param set /nova_arm_position_controller joint_limits.J1.has_velocity_limits true
./result/bin/ros2 param set /nova_arm_position_controller joint_limits.J1.min_velocity -0.1
./result/bin/ros2 param set /nova_arm_velocity_controller joint_limits.J1.has_acceleration_limits true
./result/bin/ros2 param set /nova_arm_velocity_controller joint_limits.J1.max_acceleration 0.1
```
at runtime there is a separate copy of the params for ik and fk. It may say you need to set the min_xxx before has_xxx_limits but I think that's a false error message. Whenever it changes from ik to fk there will be stuff printed saying what the limits for each joint are.

there is a bug! I'm not sure if it is in this code or elsewhere. Upon switching from fk to ik, the arm's position is set to all joints zero, but it is quickly moved back to the ik target located at where the arm was before this switch. You can see this happening at the end of the second video below. I feel like this is something that only will happen in simulation but its hard to know for sure.

## Screenshots

split into two because it was too big
[arm_limits_p1.webm](https://github.com/user-attachments/assets/a89bde92-b7ab-4b56-b259-36bbbb102eb4)
 
[arm_limits_p2.webm](https://github.com/user-attachments/assets/d24fb5ba-b31d-4ffd-8000-7b452f58a036)

## Steps to Test

In 4 terminals:

1. `./result/bin/rviz2`
2. `./result/bin/ros2 run joy joy_node`
3. `./result/bin/ros2 run teleop_arm_joy teleop_arm_joy_node --ros-args --params-file /home/nova/nova/src/ros/rover/teleop_arm_joy/config/ps5.config.yaml` (substitute ps5 depending on what controller you have)
4. `./result/bin/ros2 launch auto_bringup mock.launch.py arm:=true`

in rviz:
1. set fixed frame to base link
2. add -> by display type -> robot model
    1. in robotModel, set Description topic to /robot_description
3. add -> by display type -> tf
     1. in tf, set Filter (whitelist) to "target" (so you can just see the ik target)

## Memes

before stopping the weird acceleration limiting:
[Screencast From 2025-04-12 22-24-25.webm](https://github.com/user-attachments/assets/23409229-f840-407a-b5f4-16ddfe76d5d1)


BaileyChessum (approved):
Looks really good! Nice use of standards. Should we open up a distinct (low priority) issue for jerk limits?

Redecorating:
> Should we open up a distinct (low priority) issue for jerk limits?

I've made one here https://github.com/MonashNovaRover/nova/issues/424

Do you want to merge in the twistmapper and then I'll rebase this on top of arm-master after that merge? Both prs touch nova_arm_controller.cpp

PR #422, opened 2025-04-13
Refactor URC science page

### LINKED ISSUE #419: Remove Platform Widget
The platform widget needs to be removed from the GUI since we are now using two drills. However we do need to keep the mixer toggle.

What needs to be done:
- [x] Move the mixer toggle to the bottom right of the image where other toggles are (as a third toggle in that row), can make this look nicer if time permits but not necessary and idk what could be done.
- [x] Remove the Platform widget in the middle left.

Image

FelicityMatt:
## Description

Refactored the science GUI page to remove a widget we will no longer use, as well as make the page more flexible.

Functionality should not have changed, it's just updating the css layout.

## Screenshots

Before:
Screenshot From 2025-04-13 09-53-12

After:
Screenshot From 2025-04-13 10-07-25


TJ-Lee-M (approved):
Looks good :)

PR #423, opened 2025-04-13
Onshape to Robot update + Pybullet removal

RohitPilakkat:
Updated onshape to robot to v1.5.7.

Also removed pybullet. This will completely disable the ability to run `onshape-to-robot-bullet` for testing models (We don't really use this, or have a real need for it) as GCC14 has a lot of issues with it now?

## Steps to Test

Run urdf-tool.


V01DBREAKER (approved):
LTGM

Issue #371, opened 2025-04-05
Stop hijacking `auto_bringup` for new armware


BaileyChessum:
We should really mova all the launch files we use for new armware to `nova_bringup`, or an `arm_bringup`, and let auto be.

  
  Tags for Notion Integration:
  tag:arm;tag:rover

vic-bart:
? wth have y'all been doin up in auto_bringup

V01DBREAKER:
Seems like the following has been modified:
- [ ] launch/mock.launch.py - an entirely new file
- [ ] launch/gazebo.launch.py - minor change when launching urdf.launch.py from this (addition of arm param)
- [ ] launch/control.launch.py - adds 4 nodes that spawn arm controllers
- [ ] launch/urdf.launch.py - joint state publisher and arm params
- [ ] gz_bridge.yaml - arm gazebo topics
- [ ] controllers.yaml - arm controllers
- [ ] old_controllers.yaml - more arm controllers

Note:
We should keep how we spawn/run the joint_state_broadcaster across auto and arm. In auto we use a spawner in control.launch.py and use it to broadcast dynamic joints that aren't encoded along side our encoded joints.

Redecorating:
has been moved to arm_bringup

PR #425, opened 2025-04-14
[WIP] Merge `arm-master`

BaileyChessum:
## Description

**DO NOT MERGE**

This is to allow hydra to build and cache arm code for faster development! However, we need to first resolve merge conflicts.

After this, we can consider this for merging after resolving #371, and ensuring drive control code can be run *without* arm


BaileyChessum (commented):
As far as I can reasonably currently test, I believe this won't break anything

PR #426, opened 2025-04-15
Fallback for NaN reference interfaces in nova_arm_controller

BaileyChessum:
…ontroller is NaN


Redecorating (approved):
i remember seeing something like this in the twistmapper pr. did i accidentally lose that when i was merging joint limits?

BaileyChessum:
I think so, otherwise I messed it up when rebasing

PR #431, opened 2025-04-17
Cleanup object detection in preparation for URC

### LINKED ISSUE #287: Move cube_localiser.py to nav2_autonomous folder to replace cube_localisation module.
Since we are now using cube_localiser.py, to localise detected cubes from our YOLO model, we should archive the old code in cube_localisation (under nav2_autonomous) and replace it with cube_localiser.py

### LINKED ISSUE #401: Clean up cube_localiser and related yolo_ros things
This should be done so that when we use yolo_ros in sim, we don't have to support two message types in cube_localiser.

The ideal simulation setup would be:

Gazebo Cameras -> yolo_ros -> yolo_msg republisher -> cube_localiser -> tf

Whereas IRL would look like:
OAK Camera -> cube_localiser -> tf

Please also modify yolo.launch.py so that there is separate simulation vs irl modes.
Also delete "3d mode" or spatial mode from cube_localiser as we would only use cube_localiser when the object detection publisher is publishing 2d bounding boxes.

V01DBREAKER:
This PR moves cube_localiser to nav2_autonomous, archives the code there (still being used as an example) and generalises the variable names and preps the node for URC object detection.
This PR also deletes all the old unused models, please check i haven't deleted anything we still need.

Addresses #287 and #401 

Test using:
`ros2 launch auto_bringup yolo.launch.py`

The main thing is that this hasn't broken anything else in our repo, if it breaks object detection thats fine for now.
I just wanted to clean up the mess i made prior to ARCh

Our rover in the future:
https://youtu.be/fyZZaDIKCpg?si=feXrvzuyS8Vvc8ND


vic-bart:
Code lgtm, 10/10 meme (a rover isn't autonomous until it can fight for it's own freedom :fist_raised:), just needs testing and good to merge.

V01DBREAKER:
image
After some small fixes, it is now working with auto stack 100%

PR #432, opened 2025-04-18
URC NIR Probe Widget

### LINKED ISSUE #403: URC NIR Probe Widget
We are now going to use the NIR Probe for URC and need to create a widget (and maybe page) for it. We should be able to use the same NIR probe ros node with the same service and topic subsciption.

Currently there are some things hard coded within the NIR probe, I did my best to make most of it not but there are likely some Water/Ice reference within. The first step is to abstract these out and pass labels into the function component instead. Functionality for ARC should not be affected.

Most of the widgets will be able to be reused, including the `NIRProbeOutPutWidget`, `NIRProbeLEDWidget` and the `NIRProbeFileTable`. These can be reused as well as the site selector.

The new GUI section for the NIR Probe should have the following functionality:
- [ ] Take readings from the NIR Probe ( `NIRProbeLEDWidget` + `NIRProbeOutPutWidget`)
- [ ] View these readings and have the option to delete (`NIRProbeFileTable`)
- [ ] View an aggregated results view with the average reading from each site (new functionality).
  - With the ARC implementation we are only able to view the stats of one site at a time
  - For this task the main purpose is to be able to compare different sites with each other.
  - One way to do this is to do something similar to the `NIRProbeFileTableWidget` with a different `NIRProbeCalcTable`.

also while you're here it would be great to add a `clear` button to the `NIRProbeFileTable` to clear a specific site.

FelicityMatt:
## Description

Create a GUI section for NIR Probe use at URC. We don't need any calibration, just a way to compare readings from different sites.

Fixes: #403 

## Changelogs

- Removed any water/ice references within NIR Probe widgets
- Created a URC NIR Probe widget

## Screenshots

image

## Memes

image

### DIFF
--- a/src/ros/nova-gui/nova-gui/src/components/NIRProbe/ARCNIRProbeWidget.tsx
+++ b/src/ros/nova-gui/nova-gui/src/components/NIRProbe/ARCNIRProbeWidget.tsx
@@ -13,7 +13,7 @@ interface INIRProbeWidgetProps {
 }
 
 /**
- * View of all NIR Probe widgets
+ * View of all NIR Probe widgets for ARCh
  * @constructor
  */
 const ARCNIRProbeWidget: React.FC<INIRProbeWidgetProps> = () => {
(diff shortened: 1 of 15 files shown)


TJ-Lee-M (approved):
lgtm :)

BaileyChessum (approved):
lgtm! Glad we're keeping the ARC one

PR #434, opened 2025-04-18
Banksia Arm MoveIt2 Kinematics Plugin

BaileyChessum:
## Description

I implemented `kinematics::KinematicsBase` with Keenan's IK solution to move our IK logic into a reusable piece of code inside of the twistmapper. This will hopefully be the first step to making the IK controller redundant.

We need to make the control system modular, so that future robotic arm payloads with different IK solutions can be supported.  
To do this, we can load the kinematics used in twistmapper as a plugin with the `kinematics::KinematicsBase` type.

This is the same plugin used by MoveIt2, promoting compatibility with MoveIt2.

Note, some member functions remain unimplemented, or with temporary implementations. This likely means it won't work perfectly with MoveIt2 currently.

## Changelogs

- Add the `banksia_kinematics_plugin` package, exporting `banksia_kinematics_plugin::BanksiaKinematicsPlugin : public kinematics::KinematicsBase`

## Steps to Test

Not yet testable, as we don't have anything to consume the plugin.


BaileyChessum:
#440 now encapsulates this.

PR #440, opened 2025-04-20
Make `nova_twistmapper` use a MoveIt2 kinematics plugin

BaileyChessum:
## Description

This is a complete refactor of `nova_twistmapper` and how we do IK.

Superseding #434, we remove `nova_ik_controller` from the Twist IK teleoperation controller chain, making `nova_twistmapper` interface with `nova_arm_controller` directly using a custom kinematics solver plugin `banksia_kinematics_solver/BanksiaKinematicsSolver`. 

I also took the effort to make the controller much nicer, moving away from slop.

I will write more soon, but I need to go for now.

## Steps to Test

Same steps as previous PRs. Will add explicitly soonish


src/ros/rover/auto_bringup/launch/control.launch.py
```diff
@@ -47,12 +55,8 @@ def launch_setup(context, *args, **kwargs):
...
             package='controller_manager',
             executable='spawner',
-            arguments=['nova_ik_controller', '--inactive']
-        ),
-        Node( # TODO: only when arm is enabled
-            package='controller_manager',
-            executable='spawner',
-            arguments=['nova_twistmapper', '--inactive']
+            arguments=['nova_twistmapper', '--inactive'],
+            parameters=[{'robot_description': robot_description_content}]
```
Redecorating: I've been trying to use this for giving the nova_arm_controller node the urdf, I think it doesn't work (so your code falls back to getting it from the /robot_description topic) because this gives the controller_manager spawner node the params, but doesn't actually give the params to the spawned nova_twistmapper node.
```
        Node( # TODO: only when arm is enabled
            package='controller_manager',
            executable='spawner',
            arguments=['nova_arm_velocity_controller', '--switch-timeout', '10', "-p", "/tmp/launch_params_spi6omei"], #, '--inactive']
        ),
```
this sort of thing seems to work, but I'm just hard coding the params file to one that was previously generated that I know has the urdf in it.

BaileyChessum: I did it like this because its how MoveIt2 does it. I have no idea how to get it to work through launch files, though. Forgot to remove these changes to the launch file.

MoveIt2 doesn't have the topic fallback, though.

BaileyChessum: I've gone ahead an removed this for now, relying on retrieval from a topic

Redecorating (approved):
works on my machine :+1:

PR #451, opened 2025-04-23
nova_arm_controller self collision checks

### LINKED ISSUE #429: Arm self collision checks (stolen from MoveIt2)
Even with simple joint limits, the arm can still self intersect, which is very risky.

We need some way to ensure the arm can't hit itself. This would either be making functions for joint limits that depend on the state of other joints, or by simulating colliders in the arm, checking for collisions between joint linkages, and rejecting joint positions that self intersect.

This would normally be a lot of work to implement. However, I think we can steal collision checks from MoveIt2

## Stealing self intersection checking logic from MoveIt2 

We'll use parts of MoveIt2 to check for self intersections when trying to reach a position, and preventing the rover from attempting to move to that position.

(I think in `nova_arm_controller` to start with)

We can load the model of the rover using utilities from moveit2 as a `moveit::core::RobotModelPtr`. A `planning_scene::PlanningScene` then acts as a digital twin we can set the joint states of and request collision checks from. 

```cpp
// Parse URDF
std::string urdf_str = get_urdf_string_from_robot_description_topic_or_something_else();
auto urdf_model = urdf::parseURDF(urdf_str);
if (!urdf_model) {
   // Failed to parse model
}

// Create the robot model from the URDF
moveit::core::RobotModelPtr robot_model = std::make_shared<moveit::core::RobotModel>(urdf_model, nullptr);

moveit::core::RobotModelPtr kinematic_model = robot_model_loader.getModel();
planning_scene::PlanningScene planning_scene(kinematic_model);
```

Once we have a `planning_scene::PlanningScene` actually check for self intersection by requesting and modifying the current `moveit::core::RobotState`. 

Normally we would use a moveit2 move group, and apply current joint angle values to a move group inside of the `moveit::core::RobotState`:

```cpp
moveit::core::RobotState& current_state = planning_scene.getCurrentStateNonConst();
current_state.setJointGroupPositions("name_of_move_group_containing_all_the_joints_in_order", proposed_joint_values_vector);
```

This would normally be created from an SRDF in moveit2. I don't know if we can do this. 

So, we can either set joint values directly without a move group:

```cpp
moveit::core::RobotState& state = planning_scene.getCurrentStateNonConst();

state.setVariablePosition("J1", j1_state_pos);
state.setVariablePosition("J2", j2_state_pos);
// ... etc
state.setVariablePosition("J6", j6_state_pos);

// OR

std::map<std::string, double> joint_values = {
  {"J1", j1_state_pos},
  {"J2", j2_state_pos},
  // ... etc
  {"J6", j6_state_pos},
};
state.setVariablePositions(joint_values);
```

Once we have set the state of the virtual `moveit::core::RobotState`, we can actually perform a collision/self intersection check on it
```cpp
collision_detection::CollisionRequest collision_request;
collision_detection::CollisionResult collision_result;

planning_scene.checkSelfCollision(collision_request, collision_result);

if (collision_result.collision) {
    // Handle the self-collision. Reject this target state?
}
```

Redecorating:
probably need to redo the srdf stuff to not be hard coded

this makes nova_arm_controller reject movements to self colliding positions in both velocity and position control mode. I *think* bailey's twistmapper pr does this for position control but hopefully this is still useful as it applies to velocity control as well.

## Description

i make a new joint limiter interface that prevents self collisions and then use that on the arm

i feel like we should have some sort of shared nova library so that different ros nodes can use the same code (eg this and twistmapper for collision code, the ros2 controllers for speed limiting code, etc)

## Screenshots

my gnome screen recorder keeps crashing :(

## Steps to Test

ros2 launch auto_bringup mock.launch.py arm:=true

## Memes

me when my gnome screen record keeps failing to start: :(
me when i spend a lot of time trying to get the urdf and there's already a provided function for it in the controller base class: :(


src/ros/rover/controllers/nova_arm_controller/src/self_collision_limiter.cpp
```diff
@@ -0,0 +1,165 @@
...
+  number_of_joints_ = joint_names.size();
+  joint_names_ = joint_names;
+  //joint_limits_.resize(number_of_joints_);
+  node_param_itf_ = param_itf;
+  node_logging_itf_ = logging_itf;
+
+  this->urdf_str = robot_description;
+  return this->on_init();
+}
+bool SelfCollisionLimiter::on_configure(const joint_limits::JointLimitsStateDataType &) {
```
BaileyChessum, line 77: missing whitespace

src/ros/rover/controllers/nova_arm_controller/include/nova_arm_controller/self_collision_limiter.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+private:
+  std::string construct_srdf_fallback_string(const urdf::ModelInterfaceSharedPtr &urdf_model, std::string joint_group_name);
+
+  std::string urdf_str;
+
+  moveit::core::RobotModelPtr robot_model;
+  planning_scene::PlanningScene *planning_scene;
+  collision_detection::CollisionRequest collision_request;
+  collision_detection::CollisionResult collision_result;
```
BaileyChessum: in other places we use the google c++ style guide (I think?), which appends an underscore to the end of member variable names. We should probably do that here for consistency

src/ros/rover/controllers/nova_arm_controller/src/nova_arm_controller.cpp
```diff
@@ -45,6 +46,16 @@ controller_interface::CallbackReturn NovaArmController::on_init()
...
+  
+  if (!this->joint_limiter.init(params_.joint_names, get_node())) {
+    RCLCPP_ERROR(get_node()->get_logger(), "Failed to init joint limiter");
+    return controller_interface::CallbackReturn::ERROR;
+  }
+
+  if (!this->collision_limiter.init(params_.joint_names, get_node(), get_robot_description())) {
+    RCLCPP_ERROR(get_node()->get_logger(), "Failed to init collision limiter");
+    return controller_interface::CallbackReturn::ERROR;
+  }
```
BaileyChessum, line 58: This seems really clean, I'm a big fan

src/ros/rover/controllers/nova_arm_controller/src/self_collision_limiter.cpp
```diff
@@ -0,0 +1,165 @@
...
+<!-- I think some are because the back half of the closed loop isn't represented right -->\n\
+<disable_collisions link1=\"l1\" link2=\"l2cam\" reason=\"Default\" />\n\
+<disable_collisions link1=\"l1prime\" link2=\"l2\" reason=\"Default\" />\n\
+<disable_collisions link1=\"l1\" link2=\"l1prime\" reason=\"Default\" />\n\
+<disable_collisions link1=\"l1\" link2=\"l2prime\" reason=\"Default\" />\n\
+<disable_collisions link1=\"j5\" link2=\"l2cam\" reason=\"Default\" />\n\
+<disable_collisions link1=\"j6\" link2=\"l2cam\" reason=\"Default\" />\n\
+<disable_collisions link1=\"j5\" link2=\"l1prime\" reason=\"Default\" />\n\
+<disable_collisions link1=\"j6\" link2=\"l1prime\" reason=\"Default\" />\n\
+    \n";
```
BaileyChessum, line 55: We could potentially also achieve this with modifying the allowed collisions matrix, like in the twistmapper. However, for now if it works I think this is good

BaileyChessum (approved):
Looks good!

Redecorating:
> Looks good!

thanks Bailey, I will tidy those things up once I get through some assignments.

PR #454, opened 2025-04-25
Fix load_waypoint.py race condition

tytia:
## Description

There was a race condition caused by check_nav_status checking self._goal_handle before it is set by the first send_goal_async call. This is now fixed by skipping the nav status check if self._goal_handle is not set.

## Memes

image


vic-bart:
code lgtm! Added docstrings to the files. Tested and works :partying_face:

PR #458, opened 2025-04-25
GPS Localisation in Sim

V01DBREAKER:
## Description

Adds GPS localisation in sim
**Current breaking changes to auto stack (nav2)**


vic-bart:
GPS localisation in sim implemented, looks like it works when driving straight and making tight turns, code quality is good, lgtm :+1: 
Note that while visual odom is still being published, it is no longer being used.

PR #459, opened 2025-04-25
AUGERS

### LINKED ISSUE #408: AUGERS
Yes augers plural.

URC Science now has TWO augers, so yeah we need to implement control for these.

What needs control:
- Both augers are on a thing that actuates
- Each auger actuates individually as well

(total three actuations)

Ideally we would like to keep our existing single auger code and somehow incorporate the existing code into a new design.

Another issue that we face is that we now have three things that actuate up and down and that we would ideally like to use a single joystick for (we also have an analysis arm). We would also like to maintain backwards compatibility with ARCh and the old URC design.

My thinking/approach:
- update analysis arm to accept a parameter that determines which joystick it uses (currently it only uses the right one) and then this can be used with the left joystick in the launch file.
- launch two auger scripts with CAN ids altered in the launch file.
- move auger anti/clockwise buttons to ones at the base of the joysticks.
- add functionality to use the three thumb buttons at the top to distinguish which actuation is "active"

** Alternate idea **
Alternatively you could have both drills on one joystick and both analysis arm (code) on another joystick and just implement a way to distinguish which is "active"

That is how I would approach it, but if you have a different better approach go for it, just try to keep the nodes as flexible as possible and not affect other functionality.

FelicityMatt:
## Description

Added capabilities to have two augers and two analysis arms each controlled by a single joystick respectively.

## Changelogs

- added a new base class `ActivatedJoystickControllerNode`
- updated the URC launch file to use these
- renamed the auger and analysis arm files to not include urc.

## Screenshots

image

## Steps to Test

I have tested it with the joysticks as it worked as expected, also tested the arc launch file and it also worked without any changes needed.

## Memes

image


TJ-Lee-M (approved):
yayy :))

PR #471, opened 2025-05-01
Init arm controller position command interfaces on activation

BaileyChessum:
## Description

Initializes the arm controller position command interfaces on activation to be equal to the state interface values.

This should hopefully fix the jumping seen when switching between IK and joint space on the sim.

## Changelogs

- Make limits on arm more lenient
- Init arm controller position command interfaces on activation


Redecorating (approved):
LGTM

PR #472, opened 2025-05-01
Activated Node GUI

### LINKED ISSUE #408: AUGERS

FelicityMatt:
## Description

Add a camera component widget that describes the current status of tracked activated nodes.

fixes: #408 

## Changelogs

- publish ros messages from each activated node with its current status
- display this data in the GUI

## Screenshots

Screenshot From 2025-05-01 16-33-34

Screenshot From 2025-05-01 16-22-39

### DIFF
--- a/src/ros/nova-gui/nova-gui/src/components/ActivatedNodeWidget/ActivatedNodeButton.tsx
+++ b/src/ros/nova-gui/nova-gui/src/components/ActivatedNodeWidget/ActivatedNodeButton.tsx
@@ -15,12 +15,12 @@ export interface ActivatedNodeButtonProps extends ButtonProps {
 export const ActivatedNodeButton: React.FC<ActivatedNodeButtonProps> = (props: ActivatedNodeButtonProps) => {
     return (
       <Overlay
-        overlay={props.isLocked ? <div className="flex flex-col justify-center">
+        overlay={props.isLocked && <div className="flex flex-col justify-center">
           <Lock/>
-        </div>: undefined}
+        </div>}
       >
         <Overlay
-          overlay={props.isLocked ? <div className="grow backdrop-blur-[2px]"/> : undefined}
+          overlay={props.isLocked && <div className="grow backdrop-blur-[2px]"/>}
         >
           <Button
             color = {props.isSelected ? "primary" : "default"}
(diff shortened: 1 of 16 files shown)


src/ros/nova-gui/nova-gui/src/redux/RootState.ts
```diff
@@ -23,7 +23,7 @@ import {
   IRosNovaInterfacesHydraprobeData,
   IRosStdSrvsSetBoolResponse,
   IRosNovaInterfacesBmeSensor,
-  IRosSensorMsgsBatteryState,
+  IRosSensorMsgsBatteryState, IRosNovaInterfacesActiveNodeStatus,
```
TJ-Lee-M: no new line? :(

BaileyChessum (approved):

src/ros/nova-gui/nova-gui/src/components/ActivatedNodeWidget/ActivatedNodeButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+  isLocked: boolean,
+}
+
+// A button used for displaying an active node for the ActivatedNodeWidget
+export const ActivatedNodeButton: React.FC<ActivatedNodeButtonProps> = (props: ActivatedNodeButtonProps) => {
+    return (
+      <Overlay
+        overlay={props.isLocked ? <div className="flex flex-col justify-center">
+          <Lock/>
+        </div>: undefined}
```
TJ-Lee-M: you should use a conditional `&&` here instead of a ternary :)

```tsx
props.isLocked && ( 
<div className="flex flex-col justify-center">
  <Lock/>
</div>
```

- Bailey

TJ-Lee-M (approved):
very cool :)

PR #476, opened 2025-05-04
WIP - Updating URC Cache Controls

rotaryViper:
It uses a position controller to lock the servo at 0, 90 and 179 degrees depending on the can command sent to the node.

To Do:
Merge conflict with nova_bringup/launch/urc_science.launch.py
Test can-ids of servos
Test if they really do control the cache doors


src/ros/rover/nova_interfaces/srv/CacheCommand.msg
```diff
@@ -0,0 +1,10 @@
+# Determine which drill cache should be open 
+
+# enum for the current cache
+# byte 0_DEG = 0     # Open cache to 0 degrees
+# byte 90_DEG = 1   # Open cache to 90 degrees
+# byte 180_DEG = 7  # Open cache to 180 degrees
```
FelicityMatt: why 7? should this be 2 like normal?

src/ros/rover/science/science/urc_cache.py
```diff
@@ -1,58 +1,122 @@
...
+            max_angle=self.declare_parameter(self.SERVO_MAX_ANGLE_PARAM, self.SERVO_MAX_ANGLE_DEFAULT).value,
+            positions = self.positions
         )
+        self.cache_servo.update_position(self.DEG_0)
+
+        # Only accept inputs being delivered to receiving CAN IDs
+        self.bus.set_id_filter(self.CACHE_SEND_FRAME)
 
+        ## Create SERVICE
+        self.command_service = self.create_service(CacheCommand, '/science/cache_command', self.command_callback)
```
FelicityMatt: question: we have two nodes running how will we know which cache is being commanded to move? 

personally i think we should either publish to different topics or include a field in the message type

FelicityMatt (commented):
looking good so far! i'll have a better review at another time on my laptop, but a couple things to look at in the mean time

src/ros/rover/nova_interfaces/srv/CacheCommand.msg
rotaryViper: Oops typo 🤭

src/ros/rover/science/science/urc_cache.py
rotaryViper: Publishing to different topics sounds like the wiser decision

rotaryViper: It now adds the final number of the can address to the cache_command service
'/science/cache_command_0'
'/science/cache_command_1'

src/ros/rover/nova_bringup/launch/urc_science.launch.py
```diff
@@ -25,83 +25,125 @@
 
 def launch_setup(context, *args, **kwargs):
     # parameterised canIDs
-    auger_drill_canid = LaunchConfiguration('auger_drill_canid')
-    auger_actuation_canid = LaunchConfiguration('auger_actuation_canid')
+    auger1_drill_canid = LaunchConfiguration('auger1_drill_canid')
```
FelicityMatt: you might want to rebase off of master or something, these changes should already be in master

src/ros/rover/science/science/urc_cache.py
```diff
@@ -1,58 +1,122 @@
...
+            max_angle=self.declare_parameter(self.SERVO_MAX_ANGLE_PARAM, self.SERVO_MAX_ANGLE_DEFAULT).value,
+            positions = self.positions
         )
+        self.cache_servo.update_position(self.DEG_0)
+
+        # Only accept inputs being delivered to receiving CAN IDs
+        self.bus.set_id_filter(self.CACHE_SEND_FRAME)
 
+        ## Create SERVICE
+        self.command_service = self.create_service(CacheCommand, f'/science/cache_command_{self.CACHE_SEND_FRAME[-1]}', self.command_callback)
```
FelicityMatt, line 99: where is cache send frame defined? i'm not sure it's a good idea to use a can reference as a topic name, i would pass in the number as a parameter then we can set it to whatever

FelicityMatt:
dupe of: https://github.com/MonashNovaRover/nova/pull/485

PR #483, opened 2025-05-07
nova_arm_controller: parameter to emulate position control

Redecorating:
set the broken_position_control param to true to enable this.

Even in simulation I notice some oscilations (but they settle down quickly). Depending on how it is IRL we may need to be smarter and do a PID loop or smth.


Redecorating:
im not sure if arm-master is the right base branch for this - i forgot where the latest blcmd stuff is

PR #485, opened 2025-05-08
Double Cache Control

### LINKED ISSUE #407: Cache Control
There will be two **_positional_** servos (one per drill), to control the caches. Controlling these servos will need to be updated. 

Current (single) Cache Toggle:

Image

What needs to be done:
- Update service to select which cache is open/closed
- Update ROS2 node to use positional servos for the two caches
  - each cache will have two positions to travel to (open and closed)
  - can use existing python control, see arc spinny part for an example
- Update the GUI to have toggles/buttons for the two caches with corresponding service calls.
  - might be nice to make the open/closed state a bit easier to understand rather than an unlabelled toggle
  - make adjustments to the current gui layout as necessary

FelicityMatt:
## Description

Control the positions of the two science caches.

Will still need to be tuned and CAN commands updated but general functionality works locally.

fixes: #407 

## Changelogs

- update cache control
- add selection tabs on the gui to switch between angles

## Screenshots

image

[Screencast From 2025-05-08 10-03-15.webm](https://github.com/user-attachments/assets/32f8c38d-a675-4ab6-bb18-ff2795072d69)

## Memes

aaahhhhh

crunch time


TJ-Lee-M (approved):
:+1: :)

PR #488, opened 2025-05-08
Keyboard Localiser working IRL with GUI overlay

V01DBREAKER:
## Description
Keyboard localiser detects the model keyboard and publishes a transform.
It also publishes points used on the GUI to create a visible overlay.

Importantly cameras2 is modified to have the pipeline also publish a ros image topic if the camera serial is given a ros_topic parameter. See config.yaml in `cameras2/cameras2/params`.
This is the main deciding factor on if this should be merged!

## Steps to Test
Terminal 1: Run GUI and navigate to `urc/auto-typing`
Terminal 2: `rosbridge`
Terminal 3: `ros2 run arm keyboard_localiser.py --ros-args --params-file /home/nova/nova/src/ros/rover/nova_bringup/params/arm.yaml`
Terminal 3: `ros2 launch cameras2 camera_server_launch.py`

## Memes
image

### DIFF
--- a/src/ros/cameras2/cameras2/cameras2/camera_ros_streamer.py
+++ b/src/ros/cameras2/cameras2/cameras2/camera_ros_streamer.py
@@ -7,6 +7,10 @@ This service manages a GStreamer pipeline to
 stream video footage from ros Image topics over WebRTC.
 Consult the repository README for complete setup
 instructions.
+This node can only be run when the camera stack is not being run
+    e.g for gazebo sim only
+
+There is also gstreamer pipeline classes at the end.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 NODE: camera_streamer
 TOPICS: None
@@ -54,6 +58,9 @@ from sensor_msgs.msg import Image
 
 
 class CameraStreamerService(Node):
+    """
+    This node is 
+    """
     class CameraConfiguration():
         def __init__(self, serial: str, topic: str):
             self.serial = serial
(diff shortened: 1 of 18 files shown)


src/ros/cameras2/cameras2/cameras2/camera_streamer_service.py
```diff
@@ -46,8 +47,9 @@ class CameraStreamerService(Node):
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     PACKAGE: 	cameras2
     AUTHOR(S):	Joshua Leivenzon
+    EDITED BY:  Anthony Lew
```
hacker1024:
```suggestion
    AUTHOR(S):	Joshua Leivenzon, Anthony Lew
```

They're close enough in meaning, don't sell yourself short 🙂
We seem to have a mix across the codebase, but not really many occurrences of both together as far as I can tell

Tbh I've always thought these headers to be useless when Git exists anyway but that's another conversation

src/ros/cameras2/cameras2/cameras2/camera_streamer_service.py
```diff
@@ -317,7 +319,11 @@ def _update_cameras(self, cameras: Cameras) -> None:
 
     def _create_camera_bin(self, serial: str, device_node: str) -> CameraWebRTCBin:
         camera_configuration = self._camera_configurations.get(serial, self._default_camera_configuration)
-        camera_bin = CameraWebRTCBin(
+        # Choose the gstreamer bin to use
+        # If a ros topic is present, use the split ros topic and webrtc pipeline
+        # TODO: Make this more robust and able to support any sort of pipeline in a plugin style system able to be specified in param file
```
hacker1024, line 323: Amen. This will help with CSI cameras also. Though I reckon we could benefit from introducing _multiple_ bins that are tied together, like a source-specific bin and a common processing bin or something.

Doesn't need to be done in this PR though of course.

src/ros/cameras2/cameras2/cameras2/camera_ros_streamer.py
```diff
@@ -54,6 +58,9 @@
 
 
 class CameraStreamerService(Node):
+    """
+    This node is 
+    """
```
hacker1024: Forgotten docstring?

hacker1024 (commented):
Cool stuff! cameras2 LGTM - though it would be nice to share a little more code between the two bins in the future, as they do still have quite a lot in common.

src/ros/cameras2/cameras2/cameras2/camera_ros_streamer.py
V01DBREAKER: Removed it thanks

PR #496, opened 2025-05-11
Fix the old arm

novarover:
## Description

How fix
- removing the While True: in the resolver publisher

Zeroing the arm without the zero-arm alias, and sending the can commands manually

and use `…/ros2 launch urdf.launch.py rover:=false arm_urdf_path:=/home/nova/nova/…/rover_description/waratah_arm/urdf/arm.urdf.xacro` for rviz, to launch robot state publisher at the same time

## Memes
image


BaileyChessum (approved):
I know a self approval is immoral, but we are too close to comp

PR #501, opened 2025-05-14
Update Pumps and LED ROS nodes

FelicityMatt:
## Description

Updated pumps and led ros nodes to align with current payload, including removing an unused pump action, and updating the led commands to be in line with the new canbus standards with the STM cards.

The pumps widget will be updated in a future PR where I hope to do cool stuff with it.

## Memes

image


src/ros/rover/science/science/urc_uv_vis_leds.py
```diff
@@ -48,47 +49,20 @@ class URCUVVisLeds(ControllerNode):
     
     # CONTROL COMMANDS
     # Add any CONTROL command ids here
-    LED_1_COMMAND_ON = 0x03
-    LED_1_COMMAND_OFF = 0x04
-    LED_2_COMMAND_ON = 0x05
-    LED_2_COMMAND_OFF = 0x06
+    LED_1_COMMAND = 0x04
+    LED_2_COMMAND = 0x05
+    LED_ON_COMMAND = 0xFF
```
TJ-Lee-M: It looks like on is 01 from the notion standards page?

TJ-Lee-M (changes requested):
double check the commands for turning on led1 and led2. Otherwise lgtm :)

src/ros/rover/science/science/urc_uv_vis_leds.py
FelicityMatt: I've updated! I think it did work with this but I've updated just to be certain

TJ-Lee-M (approved):
nice :)

PR #502, opened 2025-05-15
Old arm support in new armware

BaileyChessum:
## Description

Attempts to get the old arm working with the new arm software stack. 

Currently incomplete. We still need to do the following:
- [ ] Specify a kinematics plugin that works for the old arm (maybe KDL?)
- [ ] Create simple colliders for the new arm
- [ ] Create kinematics frames on the new arm


BaileyChessum (commented):
Checked over changes, its not ideal that all the hardware testing changes are getting included in arm-master through this. Otherwise, I think this is good. We just need to be cautious of changes to BLCMD hardware potentially impacting ros2_control for drive -- and how that might affect auto

BaileyChessum:
> We just need to be cautious of changes to BLCMD hardware potentially impacting ros2_control for drive -- and how that might affect auto

Theoretically, nothing should have functionally changed, but there is a risk we might have made an error

PR #503, opened 2025-05-15
Fix gps gui

FelicityMatt:
## Description

Update gui with new ros topic 

## Screenshots
published to the topic with coords and it works

Screenshot From 2025-05-15 12-10-25

## Memes

It worked on gui on master when I only changed the topic name :eyes: 

image


TJ-Lee-M (approved):
nice

PR #504, opened 2025-05-15
Update BME Sensor

FelicityMatt:
## Description

Update BME sensor to listen and publish new data.
image

Altitude still needs to be calibrated, but we will need the hardware for that so we can calibrate in america.

## Screenshots

image


TJ-Lee-M (approved):
lgtm

PR #505, opened 2025-05-15
Fixed GPS

vic-bart:


vic-bart:
lgtm :flushed: tbh im in a rush so if this breaks stuff ill fix it :crossed_fingers: :point_right::wood:

PR #506, opened 2025-05-15
Feat/360 calculator

TJ-Lee-M:
## Description
Added manual input fields for the height of statigraphy to be calculated in gui :) 
GUI also now displays the mouse coords and the vertical angle they map to above the 360cam canvas

## Screenshots
image

## Steps to Test
Play around on /urc/360cam gui (both tabs)
From testing on obects with known dimensions to verify our approach:
- Perspective good: see screenshot... measured 6.9cm height as 7cm :D
- Panorama bad: using angles from panorama canvas gave 1.2m :/ 

## Memes
image

### DIFF
--- a/src/ros/nova-gui/nova-gui/src/components/Theta360CamWidget/HeightCalculator.tsx
+++ b/src/ros/nova-gui/nova-gui/src/components/Theta360CamWidget/HeightCalculator.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect, useState} from "react";
+import React, {useCallback, useEffect, useState} from "react";
 import {useGenericStore} from "../../hooks/useGenericStore.ts";
 import {Input} from "@nextui-org/react";
 
@@ -27,7 +27,7 @@ export interface heightCalc360CamProps {
     }, [inputThetaHigh, inputThetaLow])
 
   // set input text boxes when angles are updated from canvas
-    useEffect(() => {
+    useCallback(() => {
       //setInputThetaHigh(String(props.angles[0]));
       //setInputThetaLow(String(props.angles[1]));
     }, [props.angles])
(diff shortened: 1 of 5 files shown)


FelicityMatt (changes requested):
Can we please separate this out into it's own component (it can be external or placed in both camera views) we have repeated code here which is bad coding practice, it is also a separate functionality to the canvases so should be separate, each component should have one purpose, with other components composing many other components for a larger purpose.

I also don't think this should be added to the generic store since it is such a minor field that is only used in one spot, if you were having issues with it needing to persist between panorama/perspective I would pull it out and put it in the 360 cam view since it does not rely upon either of those views and can be a separate thing. The generic store does a lot of syncing work and I worry if we add too many things there will be a large overhead and could slow things down etc.

TJ-Lee-M:
More testing results:
- 0.10m tall object, 1.15m distance
Perpsective at the zoom point just before excessive blur from distortion: 0.076m
Panorama: 0.044cm

- 1.1m tall object, 1.15m distance
Perpsective at the zoom point just before excessive blur from distortion: 0.55m
Panorama: 0.34m

- ~0.17m tall object, ~3.4m dist
Perpsective at the zoom point just before excessive blur from distortion: 0.11m
Panorama: 0.075m

- 0.455m tall object, ~3.4m dist
Perpsective at the zoom point just before excessive blur from distortion: 0.34m
Panorama: 0.26m

TJ-Lee-M:
I have added shift-click to set high angle and ctrl-click to set low angle on both canvases :)

Currently this is repeated code in each canvas file... I think I could fix this by defining onClick in the widget file instead but this would introduce a lot more props bc its gl and event stuff?? ;~; I'm hoping to find a nicer way to generalise it if I can but idk lol

But should be working as is for now at least!

FelicityMatt (approved):
i have minor comments and improvements but at this point if it works and doesn't break anything lgtm

just make sure you do a rebase and yarn build before merging

TJ-Lee-M:
-  4+1/8 " object ~ 65+3/4 " away
Perpsective at the zoom point just before excessive blur from distortion: 2.3"
Panorama:2.5"

PR #509, opened 2025-05-18
Position control and multi-turn emulation in CMDHardware

BaileyChessum:
## Description

Adds position command interfaces via velocity control, and multi-turn resolver emulation in the CMD hardware interface.

## Steps to Test

I don't yet have a way to test this :(


Redecorating (approved):
lgtm (lets gamble try merge)

PR #511, opened 2025-05-18
KDL Kinematics for the old arm

BaileyChessum:
## Description

Add the KDL kinematics plugin to ros2_control to get simple IK working with the old arm.

Notes:
- The plugin seems to timeout whenever a limit is exceeded for the URDF joint limits
- I haven't made simple colliders for the old arm yet, so it will be expensive


Redecorating (approved):
looks sane

PR #513, opened 2025-05-20
Path planner

BaileyChessum:
## Description

I made a simple path planner. However, haven't put in a way to test it yet, as I need to pack my things to go to URC.


Redecorating:
I tried to get it to load and activate, my wip is here feat/arm/path-planner-orlando

rn its hitting [moveit_2154747338.moveit.kinematics.kdl_kinematics_plugin]: Group 'nova_path_planner_joints' is not a chain but ill look into that later if nobody else does (this might be caused by me setting the parameters for nova_path_planner wrong, but otherwise it must be some difference between path planner and twistmapper so it should be fixable)

BaileyChessum (commented):
I approve of my work :)

PR #515, opened 2025-05-21
Fix/old arm

Redecorating:
make old arm resolver code use a callback instead of polling which wasn't fast enough to handle the position telemetry without delay


Redecorating:
I know this could be tidied up more but I dont have joysticks to test if i break anything so im a bit reluctant to do any more changes on top of what I tested earlier

RohitPilakkat (approved):
Looks good to me, the previous changes were a patchwork when resolvers were being strange. I didn't know notifier existed so that's cool to know. 

I think testing it on the arm first before merging is safest but changes look sane.

BaileyChessum:
it looked sane!

PR #516, opened 2025-05-23
URC Heater Widget

### LINKED ISSUE #493: URC Heater Temperature Control
The URC heaters under the shot glasses previously just had an on/off switch on the gui, however this year we need to implement a temperature control loop.

This should be exactly the same as how the kiln worked for ARC, in fact we can probably just reuse the ros2 file and the gui widget (maybe just adding some name changes via parameters).
- actually the sensor data format will be different so this probably won't work, if possible try to use/create a python control sensor for this, but don't worry mvp without is fine.

For the temperature sensor we are using a NTC (there are two NTC sensors use the one on the same board as the heaters) Information about this is in the canbus standards.

Currently they are giving us 12 bits of 0 volts -> 4095 = 3.3 volts which science will calibrate and map to temperature data. 

example data can be found in this text file: https://drive.google.com/file/d/1920ViPJ3V8uCYnuCVRcR_OlF0Gp0Mzxs/view?usp=drive_link
The last section of data in this file is us first sticking the sensor in icy sand, then room temp, then blasting it with a heat gun to see what happens :)
you can probably make some kind of super rought calibration from that data for testing if desired

It might also be good to incorporate the other IR sensor in this widget.

FelicityMatt:
## Description

Update the heater pipeline to include temperature control and publishing data from the two NTC sensors.

## Screenshots

Screenshot From 2025-05-24 09-22-25

image

### DIFF
--- a/src/ros/nova-gui/nova-gui/src/RosRoot.tsx
+++ b/src/ros/nova-gui/nova-gui/src/RosRoot.tsx
@@ -7,7 +7,6 @@ import { Toaster } from "react-hot-toast";
 import { NovaTopBar } from "./components/TopBar/TopBar";
 import { NeoSidebar } from "./components/NeoSidebar/NeoSidebar";
 import { BLCMDStatusModal } from "./components/BLCMDStatusModal/BLCMDStatusModal";
-import { RGBInputModal } from "./components/RGBWidget/RGBModal.tsx";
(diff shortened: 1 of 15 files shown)


TJ-Lee-M (approved):
looks good  :)

PR #520, opened 2025-05-24
cmd_hardware: move resolver discontinuity to +-180deg

Redecorating:
cmd_hardware: move resolver discontinuity to +-180deg

In the case of j6 specifically, this means that if you start the
hardware interface while it is within 45 degrees of the zero position
its good. Without this, if you were for example at -1 degree, it would
think j6 is at 44 degrees (because of the 4 as the gear ratio).

For other joints I don't think it matters because they don't go past +-180
degrees (and if they do in simulation, Bailey's code still keeps track of how
many revolutions they have taken regardless). The joints other than j6 don't
seem to have any issues caused by this change.

More detailed context is in the commit message.

## Memes

  
  Tags for Notion Integration:


Redecorating:
closed by https://github.com/MonashNovaRover/nova/commit/8fa9cbdd622d4257bf93a735a5a3cb3bc8cd8acc

PR #523, opened 2025-05-24
Auto Typing Sequencer

### LINKED ISSUE #367: Autonomous Typing Behaviour Sequencer
Terse name for something rather simple. 

Assume we have the #364 Path Planner and #365 Pokey Thing actions finished, in addition to an accurate keyboard reference frame in TF2 from #366. We need a ROS2 node that can put all the pieces together to type a full sentence. 

It should expose an [Action](https://docs.ros.org/en/jazzy/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Actions/Understanding-ROS2-Actions.html) for typing a string on the keyboard.

It should accept a string to type, then using the keyboard reference frame (and child reference frames for each key!), it should sequentially ask for Path Planner actions to move between keys, and Pokey Thing actions to press each key.

V01DBREAKER:
Sequencer just needs poke integration,
Has path planner teleporting to 0 bug fix (not part of sequencer) 
Needs IRL testing. 
Ready for merge to arm-master for further integration.

How to test and run:
path.mock.launch.py
rosbridge
GUI
keyboard localiser
sequencer


V01DBREAKER:
Changed branch to feat/arm-sequencer because i hate rebasing with conflicts

PR #535, opened 2025-05-26
Feat/urc bt

tytia:
## Description

URC behavior tree nodes and XML and related changes.

### Behavior Tree
- restructured nova_behavior_tree directory structure: nodes are now sorted into action, condition and archived. Archived nodes are removed from the CMake file.

### New Nodes
- PlaceSearchGoalsAction: places search goals in the shape of an equilateral triangle to search for AR tags / objects within a radius.
- UpdateURCGoalsAction: upon receiving the first trustworthy detection, it removes all other goals and inserts the detected pose, which it will keep updating from then on. This node does not have a check for update radius, as the intention is that any detection goal passed to it should be accurate.
- RemovePassedURCGoals: removes goals based on proximity and orientation, does not remove the last goal so that navigation is finished by FollowPath.
- DetectedConditionBase: abstract class which ARTagDetected and ObjectDetected inherit from, it subscribes to a detection topic given via input port detections_topic and filters detections by grouping them into clusters based on proximity and only publishing the latest goal in a cluster if there are at least min_detections detections in buffer_time seconds in the cluster.
- ARTagDetected: DetectedConditionBase<aruco_opencv_msgs::msg::ArucoDetection>, detection topic "/aruco_detections"
- ObjectDetected: DetectedConditionBase<visualization_msgs::msg::MarkerArray>, detection topic "/yolo/objects"
- ArePosesNear: checks if two poses are within a given distance, transforms them to the same frame if they are not within the same frame (copied from Nav2 rolling)
- GetCurrentPose: gets the current pose of the rover (copied from Nav2 rolling)
- GetGoal: given a goals array, get the goal from a given index.

### Modified Nodes
- SnapInCollisionGoals: removed cube-related logic, is now mostly self-reliant. It assumes that all new goals are placed with an offset and therefore calculates the toward points using that offset.

### Other
- load_waypoint.py now supports starting with a GPS coordinate given through params if waypoints.json does not exist.
- New BT navigator: URC2025Navigator, which is the same as NavigateThroughPoses except it adds a 'type' bb variable which represents the type of goal we are traversing towards (0 = Nothing, 1 = AR Tag, 2 = Object).
- added and tuned params robot_radius and goal_offset_radius under bt_navigator in nav2_urc.yaml for BT nodes to access

## Screenshots

https://github.com/user-attachments/assets/9d052c68-2b71-4bb2-b7fd-78dd051d1666

2025_urc_through_poses

## Memes

image


vic-bart:
Tested in sim = works, code reviewed, lgtm

src/ros/rover/nova_utils/nova_utils/load_waypoint.py
tytia, line 1: should we have default search radii for AR tags and objects that are overridable instead of just an overridable search radius, so there's one less thing to remember while running the file?
```python
self._search_radius = self._ar_tag_search_radius if self._type == 1 else self._object_search_radius
```

PR #536, opened 2025-05-27
Removing autonomous ankle-breaking

vic-bart:
## Description

No more ankles will be broken!


vic-bart:
tested in sim, works, lgtm

PR #541, opened 2025-05-28
Feat/magnetometer

novarover:


novarover:
tested IRL individually, not on both cameras simultaneously on rover, but lgtm

PR #542, opened 2025-05-28
Zeroing red and blue LEDs before flashing green

kuhu133:
## Description

Zeroing red and blue LEDs before flashing green

### DIFF
--- a/src/ros/rover/electronics/electronics/led_strip.py
+++ b/src/ros/rover/electronics/electronics/led_strip.py
@@ -63,6 +63,7 @@ class LedStrip(ControllerNode):
             self.set_duty_cycle(self.RED_CONTROL_ID, 0)
             time.sleep(0.05)
             self.set_duty_cycle(self.BLUE_CONTROL_ID, 0)
+            time.sleep(0.05)
 
             # Save values and start flashing (only for green for now)
             self.flash_g = request.g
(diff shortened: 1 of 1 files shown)


src/ros/rover/electronics/electronics/led_strip.py
```diff
@@ -59,6 +59,11 @@ def led_service_callback(self, request, response):
                 self.flash_timer.cancel()
                 self.flash_timer = None
 
+            # Clear all current LED states except green
+            self.set_duty_cycle(self.RED_CONTROL_ID, 0)
+            time.sleep(0.05)
+            self.set_duty_cycle(self.BLUE_CONTROL_ID, 0)
```
FelicityMatt, line 65: do we need to add another sleep after this?

kuhu133: I wasn't sure if sending the two Can commands at the same time would cause issues coz sending three did, so just thought i should add a small delay. Do you think it would be better to test without it first?

FelicityMatt: oh sorry we do need it, I meant do we need another one after the setting of blue control

FelicityMatt: ah all good I just saw that you start a timer to flash green so all good.

FelicityMatt (approved):
lgtm, merge it!

src/ros/rover/electronics/electronics/led_strip.py
kuhu133: oh truee, yep will add thatt, tyy!

PR #545, opened 2025-05-28
WIP: Arm/oldarm ik

Redecorating:
keenan's ik solution for old arm

seems to be an issue in the z coord but otherwise i think its working ok

america people feel free to push to this and rename it to not WIP if you fix the z issue :)

## Memes

  
  Tags for Notion Integration:

### DIFF
--- a/src/ros/rover/arm_bringup/launch/control.launch.py
+++ b/src/ros/rover/arm_bringup/launch/control.launch.py
@@ -41,7 +41,7 @@ def launch_setup(context, *args, **kwargs):
                 Node(
                     package='controller_manager',
                     executable='spawner',
-                    arguments=['nova_arm_velocity_controller', '--inactive'] #, '--inactive']
+                    arguments=['nova_arm_velocity_controller'] #, '--inactive']
                 ),
                 Node(
                     package='controller_manager',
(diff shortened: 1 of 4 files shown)


BaileyChessum (approved):
Looks good! I'm going to test this with master, as nothing else can break

Redecorating:
I've put a commit that makes it choose the elbow in/out solution that is closest to the current arm state on `feat/oldarmik/elbowInOut`. I'm not sure if we want this, as the solutions with j3 down have a lot more restrictions on movement as j3 can collide with the arm box, however it does fix the issue of the arm snapping to a different position when you enable ik with j3 down.

FelicityMatt:
@Redecorating what's the current status of this?

Redecorating:
if we want new armware ik on old arm we would need this. hopefully its just a matter of dropping the change to control.launch.py and moving the changes in controllers to arm/arm_controllers

Redecorating:
"the stupid temp fix" is still stupid but O(1) - i think there was an issue with the measurements or some discrepancy between where keenan's ik code considered all the joints to be and where ros considered them to be

Redecorating:
replace with https://github.com/MonashNovaRover/nova/issues/637

PR #566, opened 2025-06-08
Use GUI to Control the Rover

V01DBREAKER:
This PR creates a GUI component which is able to control the rover by sending joy commands to teleop drive.

### Testing Drive
- Run GUI and navigate to `/test/joy`. You can also try on mobile by running `gui-run --host` and navigating to your pc's ip on mobile.
- Run `gui-rosbridge`
- Run teleop_drive_joy: `launch-teleop-drive local:=True` 
- Run `ros2 launch auto_bringup gazebo.launch.py local:=True` change it to short auto mount (in rover.urdf.xacro) for faster gazebo
- Unlock the controller by the GUI and drive the rover

Buttons have been updated for latest joy message of teleop drive joy

### Testing Arm
- Same steps but use the arm joy widget instead
- Don't launch gazebo but run the arm sim instead:
  - `mros2 launch arm_bringup control.launch.py use_mock_hardware:=true rviz:=true`
  - `mros2 launch teleop_arm teleop.launch.py local:=true`

Has not been tested on rover but it should work as it runs through teleop, so if teleop works this will work.

Also added a widget blur lock so it wont start sending joy messages immediately if you navigate to this page

https://github.com/user-attachments/assets/7994dc10-3e4c-48b5-a4fa-288c33f364a2

https://github.com/user-attachments/assets/2553a91f-3f98-4424-a126-8bbeb2253124

Tags for Notion Integration:
tag:nova-gui;tag:control


src/ros/nova-gui/nova-gui/src/components/TopBar/TopBar.tsx
```diff
@@ -6,9 +6,9 @@ import {
   DropdownMenu,
   DropdownTrigger,
   Navbar,
-  NavbarBrand,
+  //NavbarBrand,
   NavbarContent,
-  NavbarItem,
+  //NavbarItem,
```
FelicityMatt: Can we remove these lines if they're no longer needed.

src/ros/nova-gui/nova-gui/src/components/DriveJoyWidget/DriveJoyWidget.tsx
```diff
@@ -0,0 +1,296 @@
...
+  
+    document.addEventListener('touchmove', preventPinchZoom, { passive: false });
+  
+    return () => {
+      document.removeEventListener('touchmove', preventPinchZoom);
+    };
+  }, []);
+
+  // return a card with all components
+  // its a big blob of code i'm sorry if anyone decides to read this
```
FelicityMatt, line 191: I don't love this but I can't really see way to go about it without being equally complicated so should be fine to leave it as is.
Another option instead of doing onPressStart/onPressEnd is instead to send a joy message on click explicitly instead of indirectly through the useEffect. But that might make things more complicated so :woman_shrugging:

src/ros/nova-gui/nova-gui/src/components/DriveJoyWidget/DriveJoyWidget.tsx
```diff
@@ -0,0 +1,296 @@
...
+  };
+
+  /* send joy messages when state changes
+     must convert state to joy message axes and buttons
+     joy msg format on xbox:
+     axes: number[-1to1] = [left_x, left_y, left_trigger, right_x, right_y, right_trigger, dpad_x, dpad_y]
+     buttons: number[0:1] = [a, b, x, y, left_bumper, right_bumper, screenshot, menu, xbox, left_stick_press, right_stick_press, upload]
+     x axes are inverted e.g right = -1
+  */
+  useEffect(()=>{
```
FelicityMatt: I think it might be wise to have some kind of locking mechanism so that this doesn't send messages unless explicitly enabled? I just don't want to be in a situation where multiple people have this open and it's causing conflicting messages or someone is already driving and then someone opens this page (maybe by accident) and suddenly joy messages are being sent and it messes with things. - this wouldn't be the joy locking system it would be completely on the gui side and prevent any and all messages being sent.

FelicityMatt (dismissed):
There's a bunch of linting issues that should be addressed as well.

I also found a couple issues:

On my phone I couldn't use the full height the joysticks properly. Is this just something on my phone or is there a way to fix already, or an accepted issue?
image


The joysticks were overlayed over the side bar and modals:
Screenshot From 2025-07-06 20-33-14

Also found this issue in the console which should be looked into:
Screenshot From 2025-07-06 20-36-36

FelicityMatt:
Just found another issue: looks like the science page isn't able to scroll down in this branch

V01DBREAKER:
I have pulled out the useful things i did in this branch #621 and #622 but I believe I need to redo this component and rethink how to implement it properly. I will rename this PR as WIP but please keep this branch and PR so i can look over the comments,

V01DBREAKER:
joystick overlayed on sidebar and linting has been fixed

src/ros/nova-gui/nova-gui/src/components/TopBar/TopBar.tsx
V01DBREAKER: top bar no longer being changed :)

src/ros/nova-gui/nova-gui/src/components/DriveJoyWidget/DriveJoyWidget.tsx
V01DBREAKER: will improve in future...

src/ros/nova-gui/nova-gui/src/components/DriveJoyWidget/DriveJoyWidget.tsx
V01DBREAKER: it is not wise to have people use this during comp etc, should only be for testing

V01DBREAKER:
image

V01DBREAKER:
Added the arm cause why not

FelicityMatt (approved):
LGTM

Tested locally in sim, it is currently hidden on a test page and locked by default so shouldn't accidentally interfere with anything.

V01DBREAKER:
My reward for helping with science PRs :D

PR #584, opened 2025-06-25
Nav2 Autonomous Launch and Parameter File Refactor

V01DBREAKER:
Tired of scrolling through the 500 line nav2 yaml file looking for the parameter to change?
Sick of the **looooooong** launch file with the composition stuff we never use? (But we should in future)

This PR refactors the launch file to consider any yaml file in the directory specified under `nav2_params:=/path/to/dir` so that the giant nav2 param file can be split up however we please!

It also removes the giant repetition in our launch file by completely removing composition. 
While this was a nice ideal to have, we ultimately never used it even though it is meant to bring performance benefits. Maybe we can attempt to bring this back in the future?

This change also makes the param folder less cluttered yay!

Meme:
https://www.youtube.com/watch?v=6-1Ue0FFrHY


vic-bart:
Cleaned up, tested all changes in sim, lgtm

PR #593, opened 2025-07-12
[WIP] replace pkgs.hostPlatform with pkgs.stdenv.hostPlatform

BaileyChessum:
## Description

This follows @rotaryViper 's idea to fix hydra after it recently starting having eval errors, replacing `pkgs.hostPlatform` with `pkgs.stdenv.hostPlatform`.

DO NOT MERGE! This is yet to be tested

### DIFF
--- a/nixfiles/ci/jobsets/devices.nix
+++ b/nixfiles/ci/jobsets/devices.nix
@@ -22,7 +22,7 @@ let
       ({ config, ... }: {
         # Cross-compilation here is not useful, as these jobs are designed to
         # populate a cache to make rebuilds on the devices themselves faster.
-        nixpkgs.buildPlatform = config.nixpkgs.hostPlatform;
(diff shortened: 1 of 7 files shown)


BaileyChessum:
yep its definitely more complicated than just that. The news article isn't relevant, and the issue occurs on both old nix hydra, latest nix hydra and lix hydra. Please see fix/hydra for some of the work I was doing there, but I am no closer to figuring out why 

**_pkgs.hostPlatform doesnt exist_** for some reason? in nixfiles/default.nix. Attempting to reproduce it manually, I find it gives the expected x86_64-linux. Running it on hydra with the declarative jobset, the usual error we see on hydra.novarover.space appears. Trying to set up a non-declarative, manual jobset to replicate it in a way where I can modify the nova-monorepo source more easily, hydra has internal errors in perl when trying to evaluate the jobset. Weird.

image

The underlying error is the same, but with different formatting in the log in lix hydra.

PR #605, opened 2025-07-13
Remove rebuild warnings

rotaryViper:
## Description

Removed evaluation errors for rebuilds.

## 

## Steps to Test

- sudo nixos-rebuild test
- Should have all rebuild warnings disabled

## Memes

  
  Tags for Notion Integration:
  tag:nixfiles


nixfiles/modules/home/desktop/browser.nix
```diff
@@ -12,6 +12,9 @@ in
       commandLineArgs = [
         "--profile-directory=nova"
       ];
+      extensions = [
+        "ddkjiahejlhfcafbddmgiahcphecmpfh" # ublock origin lite
+      ];
```
BaileyChessum, line 17: "ddkjiahejlhfcafbddmgiahcphecmpfh"?? what?? over my head dawg

nixfiles/modules/home/nova/nixvim.nix
```diff
@@ -0,0 +1,162 @@
+{ config, lib, ... }:
+
+let
+  nixvim = import (builtins.fetchGit {
+    url = "https://github.com/nix-community/nixvim";
+  });
+  cfg = config.nova.desktop;
```
BaileyChessum: should we make this something like `config.nova.nixvim` instead? So people can opt in to nixvim instead? I've known some people to already use their own nixvim setup, so don't want to cause conflicts.

nixfiles/modules/nixos/profiles/shared/default.nix
```diff
@@ -11,7 +11,14 @@ in
...
 
-    environment.variables.EDITOR = "vim";
+    environment.variables = {
+      BROWSER = "chromium";
+      EDITOR = "nvim";
+      VISUAL = "codium";
+      PAGER = "nvim +Man!";
+      GDK_BACKEND = "wayland,x11,*";
+      QT_QPA_PLATFORM = "wayland;xcb";
+    };
```
BaileyChessum, line 17: will this affect anyone who uses a different IDE or browser?

BaileyChessum (commented):
I think I don't know enough about some of the changes to be able to make a full review yet!

nixfiles/modules/home/desktop/browser.nix
rotaryViper: https://chromewebstore.google.com/detail/ublock-origin-lite/ddkjiahejlhfcafbddmgiahcphecmpfh?hl=en

The url for the adblock, its unfortunately how nixos likes its chrome extensions set 🥲

nixfiles/modules/home/nova/nixvim.nix
rotaryViper: Yep I'll change that, and I'll make individual lsps and options modules! etc colorizer or start page

nixfiles/modules/nixos/profiles/shared/default.nix
rotaryViper: Anyone using a different browsers and editors can override these options using `lib.mkForce` for their own setup. You are right though, the defaults in the current nixfiles are pretty good:

EDITOR = vim
VISUAL = codium
PAGER = less

EDITOR should be a lighter text editor.
VISUAL should be the IDE.
PAGER is the documentation reader.

I'll remove the hardcoded env variables here.

Redecorating:
maybe can we split this into two pull requests? one for fixing the deprecation warnings and stuff, and then one for adding in the theming/plugins.

rotaryViper:
Not needed anymore

PR #621, opened 2025-07-18
Remove use_local_mesh param and add param to disable auto mount for urdf

V01DBREAKER:
These were implemented in the gui-joy branch but while i fix that branch lets merge this into master.
This PR will remove the local_mesh stuff which doesn't do anything and was originally added by me without testing at ARCh.

The param to disable the cameras in gazebo will allow us to quickly launch a sim that can be used for showcasing the rover in sim. Since the auto mount cameras slow down the sim from 90% irl speed to 12% irl speed on most laptops.

Terry please check that these changes work and don't break anything.
(They shouldn't as they've already been tested extensively on gui-joy)

`ros2 launch auto_bringup gazebo.launch.py camera:=False` to test without auto mount in sim


tytia (approved):
Tested, lgtm!
image
image

PR #634, opened 2025-07-29
Generic Interfaces

FelicityMatt:
## Description

A new package for generic interfaces to be used as much as possible.

### Motivation

there is both the std_msg and example_msg interface packages that contain similar interfaces, however many of the std messages are deprecated so we should be moving away from them. example_msg would work but they have a scary message saying don't use these they should only be for test cases. :(

These messages and services also include a standard header which will be good practice to include time stamps and we can use to see latency!

## Changelogs

- add `generic_interfaces` package

## Screenshots

image

## Memes

image

Tags for Notion Integration:
tag:software;tag:rover;tag:ready_for_review

### DIFF
--- a/nixfiles/packages/ros/nova-workspace/default.nix
+++ b/nixfiles/packages/ros/nova-workspace/default.nix
@@ -57,6 +57,7 @@
 , nova-banksia-kinematics-plugin ? throw "nova-banksia-kinematics-plugin is needed, but not available!"
 , nova-waratah-kinematics-plugin ? throw "nova-waratah-kinematics-plugin is needed, but not available!"
 , gpsd-client ? throw "gpsd-client is needed, but not available!"
+, nova-generic-interfaces ? throw "nova-generic-interfaces is needed, but not available!"
 
   # Configuration options
   ## Include graphical applications in the workspace.
@@ -106,6 +107,7 @@
       nova-path-planner
       nova-banksia-kinematics-plugin
       nova-waratah-kinematics-plugin
+      nova-generic-interfaces
       reolink;
   }
 
--- a/src/ros/rover/.gitignore
+++ b/src/ros/rover/.gitignore
@@ -35,6 +35,7 @@ __pycache__/
 
 ## Pycache
 __pycache__
+*cmake-build-debug/
 
 
 ## Rope project settings
--- a/src/ros/rover/nova_generic/README.md
+++ b/src/ros/rover/nova_generic/README.md
@@ -1,0 +1,42 @@
+# Nova Generic ROS2 Packages
+
+## Overview
+
+This collection of ROS2 packages is designed to provide reusable, configurable code components that can be leveraged across multiple projects and use cases. 
+The goal is to reduce the proliferation of custom nodes and interfaces by offering generic interfaces, classes and nodes that are easily customizable through configuration files.
+
+## Current Packages
+
+- [**generic_interfaces:**](./generic_interfaces)  
+  Generic message and service definitions for simple message types, providing building blocks for various use cases.
+
+- [**python_control:**](./python_control)  
+  A number of classes that can be used to abstract away both high level logic and low level communication with the CAN bus.
+
+- [**python_control_old:**](./python_control_old)  
+  An old version of the above used for legacy scraper/tilesplacer code - TODO remove!!!!
+
+## Contribution Criteria
+
+When adding a new package, node, or component to this collection, please ensure it meets the following guidelines:
+
+- **Generic Purpose:**  
+  The package should serve a generic purpose that can be reused in a variety of contexts.  
+  ✅ Examples: `generic_interfaces`  
+  ❌ Not suitable: `ARC_science_messages`
+
+- **Generic Naming:**  
+  The name should reflect its broad applicability.  
+  ✅ Examples: `generic_can_nodes`  
+  ❌ Not suitable: `science_sensors`
+
+If your package does not align with these principles, consider whether it can be refactored to be more generic. 
+If it is inherently domain-specific or highly specialized, it is more appropriate to add it to other packages such as `science`, `chassis` or `ec`.
+
+Feel free to reach out on the `#software-help` channel if you need help assessing or refactoring your code to make it generic!
+
+## Development Guidelines
+
+- Aim to maximize code reuse from this collection to maintain consistency and reduce duplication.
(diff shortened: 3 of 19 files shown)


TJ-Lee-M (approved):
lgtm :)

vic-bart:
This code is not ready to merge:
1. This PR gives an opportunity to discuss our design principles for the upcoming design cycle, and should be explored.
2. nova-generic-interfaces (or generic_interfaces) fails to build due to missing imports. It is an easy fix, but its existence implies no one attempted to evaluate nova-workspace before approving this PR, and that the code has not been reviewed to an acceptable standard.
3. nova-interfaces (or nova_interfaces), which generic_interfaces intends to replace, still exists which implies its removal and replacement with generic_interfaces has not been achieved nor tested.

vic-bart:
(copying from Slack to preserve it)
For anyone interested in this, here's a summary:
- May 2020, std_msgs was [deprecated](https://github.com/ros2/common_interfaces/pull/116) on ROS2 (but continued to exist to avoid breaking existing use cases in the ROS2 ecosystem)
- Sep 2023, an [discussion](https://discourse.openrobotics.org/t/discussion-standardizing-ros-interfaces/33693/3) was held on the Open Robotics forum that makes a good case against general msg and srv definitions
  - "The core idea is that a service like GetFloat shouldn’t exist because it’s unclear what the float is. A temperature in Celsius? Fahrenheit? The number of casualties?"
  - I back this point; a move away from generic to specialised msgs/srvs introduces a self-documenting code aspect to our stack.

- nova_generic/README.md
"maximize code reuse from this collection to maintain consistency and reduce duplication."
  - @FelicityMatt Is it necessary to maximise code reuse of similar msgs/srvs given A) how declarative (easy-to-create) msgs/srvs definitions are, and B) how every specialised rover folder for chassis, auto, science, arm, already has an X_interfaces package?

FelicityMatt:
To respond @vic-bart:
1. Yep! I raised this PR for discussion, I brought it up within the software meeting and no one had strong opinions about it.
2. That's weird, I was using this package for my generic nodes, and it was working fine and would've run `ws-build` before raising the PR - I'll have a look.
  I do agree though that we need to either improve or enforce a good review process - that should have been picked up in a review, I might write something about this in our github processes and highlight it to the recruits - maybe a guide or checklist for reviewers to follow - maybe Anthony can add something to his workshop.
  I also want to look into integrating with hydra to add the hydra build status to the PR to ensure we always merge working builds.
4. Removing nova_interfaces would be a non-trivial undertaking, things need to change and be tested within the GUI and every package, plus many new interface packages. My intention was to define something to move to and then slowly deprecate nova_interfaces. I would also like to rework a lot of science so it feels a bit pointless to move everything and then change everything again, but we can do that for clarity.

---

Regarding the generic interfaces discussion:
- Yeah I'm happy to move to more declarative messages, I made this after our discussion to move towards more generic messages, but it does make sense to have meaningful messages.
- The move to more interface folders will help the initial problem of nova_interfaces becoming too bloated, so I'm fine to do that.
- I use some of these in my generic can nodes, so we may need to keep and just add to the package as needed instead of anticipating needs, instead of of generic interfaces for the whole team, it can be interfaces for generic nodes.

vic-bart:
1. Awesome!
2. I checked it out, and looks like right before you opened the PR, you cleaned up the branch and (I assume by mistake?) deleted some imports to the propagatedBuildInputs for nova-generic-interfaces. Just add them back in and I or Tash can (thoroughly) review it.
3. That's a great point, the meme mislead me :laughing:. Can you add the following (or equivalent) to nova_interfaces/README.md 
    > `nova_interfaces` is in the process of being deprecated. Avoid using `nova_interfaces` in future developments, and instead use subteam-specific (e.g. `science_interfaces`) or `generic_interfaces`.

TJ-Lee-M:
Yes very good points - I think a PR Review checklist is something we should add to the PR template - We can put in the default instructions to test part so it can be added to / modified as appropriate when the PR is raised.

Outside of the template we can just emphasize the steps in the PR process and what they mean - ie when to post in the #software-pull-requests, when to approve PRs, etc, particularly for big changes (what indicates working/tested code vs approving theoretically vs ready to merge ie working AND approved by wider team / relevant people ).

Solving that ambiguity I think can avoid causing concern for anyone missing context and make sure steps aren't missed in the review process.

TJ-Lee-M:
And tyvm Victor for transferring the Slack discussion! I think we should also emphasize when discussion belong on Slack/Notion/GH to minimize lost information to help with this :)

TJ-Lee-M:
I've updated the [Notion processes documentation](https://www.notion.so/Processes-23b45d15ad2c419f8e662590078f86f7) from the recruit hub to explicitly define expectations for PR etiquette. Lmk if you have any notes or anything to add (maybe in a Slack msg I dont want to clutter this PR with this any more 😅 ) @FelicityMatt @Chetan-Edupalli @tytia - After these are formally introduced to the team I want to also write them down in a more obvious place as well to remind people of them.

image
https://www.notion.so/Processes-23b45d15ad2c419f8e662590078f86f7

image
https://www.notion.so/Git-b7ed95959aab4cb3808d4c260d59dc3c

FelicityMatt:
Our ways of working were a bit messy and all over the place before so I moved everything to a new page and created sync blocks so we can put it lots of places but it is defined once:
https://www.notion.so/Git-Ways-of-Working-243b71396171809baca2f41142582421
in the github page: https://www.notion.so/GitHub-f36a93aa7e174859a2e7a39338feb15f

FelicityMatt:
also I think I'll close this PR so no need to fix it

vic-bart:
Wait, we're not merging this? What happened? Is this not a much-needed PR?

FelicityMatt:
I thought we were going to move to defining our own interfaces with descriptive names in each areas interface package? We don't need this then. -- if we're just going to use this for generic packages I'll create it when I create some generic nodes for it to be applied.

FelicityMatt:
I have created a decision record page on notion that we can use to record our discussions: https://www.notion.so/Use-of-ROS2-Interfaces-245b7139617180ffbba9df37ccc6b1be

FelicityMatt:
I'm going to close this, I think it might be better to create some more semantic message types, maybe like "Effort" or "command" types that are still generic but with more meaning, I will look into it with python control v2

PR #646, opened 2025-08-07
Feat/auto cartographer

Matthewheekh:
## Description

Added required cams to URC and ARC autonomous camera pages
Separated auto features from cartographer and added input option for bottom overlay components
Added drawer functionality for bottom overlay to completely hide/unhide content
Reworked animations
Reordered components in bottom overlay
Redesigned auto status component
Added optional label inputs for future requirements

## Screenshots

https://github.com/user-attachments/assets/bc32e0ae-0a3f-4f3f-b69b-24e5c08d507f

https://github.com/user-attachments/assets/c5cb1d0a-2b11-4962-9b40-c5f79ef65b59

## Steps to Test

**Test cartographer in regular pages vs autonomous page**

Routes: 
- /urc/cartographer
- /urc/autonomous-navigation

**Test autonomous camera pages in URC & ARC**

Routes: 
- /urc/cameras/autonomous-navigation
- /arc/cameras/autonomous

Tags for Notion Integration:


src/ros/nova-gui/nova-gui/src/components/maps/Cartographer/components/BottomOverlay.tsx
```diff
@@ -14,27 +14,27 @@ import {
...
 interface BottomOverlayProps {
   mapTile: MapTile;
   setMapTile: (tile: MapTile) => void;
   deletePoint: (point: MapPoint) => void;
+  bottomOverlayComponents?: React.ReactNode[];
 }
 
-export const BottomOverlay : React.FC<BottomOverlayProps> = ({mapTile, setMapTile, deletePoint}) => {
+export const BottomOverlay : React.FC<BottomOverlayProps> = ({mapTile, setMapTile, deletePoint, bottomOverlayComponents = []}) => {
+  const [overlayVisible, setOverlayVisible] = useState(false);
```
FelicityMatt: can we set the default to be true? I think we will want to use it at the start and then toggle off later.

src/ros/nova-gui/nova-gui/src/components/maps/Cartographer/components/BottomOverlay.tsx
```diff
@@ -43,149 +43,162 @@ export const BottomOverlay : React.FC<BottomOverlayProps> = ({mapTile, setMapTil
...
+            borderBottomRightRadius: "0px",
+            borderTopLeftRadius: "8px",
+            borderTopRightRadius: "8px",
+          }}
+        >
+          {overlayVisible ? <ChevronCompactDown /> : <ChevronCompactUp />}
+        </Button>
+      </div>
+      {overlayVisible && ( 
+        <motion.div
```
FelicityMatt, line 74: can we animate this when it pops down as well? I was a bit jarred when it only animates one way.

[Screencast From 2025-08-11 12-06-04.webm](https://github.com/user-attachments/assets/8c364824-24f7-4fbf-8511-dd1946b294f6)

FelicityMatt: Also as you can see when you load the map there is a bar of no mapped region when the bottom overlay is gone. It might be nice to adjust the bottom depending on whether the bottom overlay is present or not (ie all the way down when no overlay and where it is now when there is overlay)

src/ros/nova-gui/nova-gui/src/components/maps/Cartographer/components/CartographerGoalModal.tsx
```diff
@@ -31,16 +31,29 @@ import {
 import { SortablePoints } from "./SortablePoints.tsx";
 import { useBifrost } from "../../../../redux/actions/bifrost/useBifrostAction.ts";
 import { RosService } from "../../../../ros/services/rosService.ts";
-import { GoalType, MapPoint } from "../../../../redux/models/CartographerState.ts";
+import { GoalType } from "../../../../redux/models/CartographerState.ts";
 import { IRosNovaInterfacesCartographerCommandRequest } from "../../../../ros/rosTypes.ts";
+import { useSelector } from "react-redux";
+import { RootState } from "../../../../redux/RootState.ts";
+
+interface Point {
```
FelicityMatt: maybe this could extend from the MapPoint type? then we won't have duplicate types since it looks like this just has one extra field

src/ros/nova-gui/nova-gui/src/views/shared/CamerasPage/CameraPageConstants.tsx
```diff
@@ -97,6 +97,13 @@ const driveCams = [
   CameraSerials.DRIVE_TELEMETRY,
 ]
 
+const autoCams = [
+  ...mastCams,
+  ...driveCams,
```
FelicityMatt: can we remove mast and drive cams here, this section is for auto-specific cameras and then we merge them together later

src/ros/nova-gui/nova-gui/src/views/shared/CamerasPage/CameraPageConstants.tsx
```diff
@@ -174,7 +181,7 @@ export const space_resources_views: CameraView[] = [
 
 export const autonomous_views: CameraView[] = [
   {
-    cameraSerials: [...mastCams, ...driveCams],
+    cameraSerials: autoCams,
```
FelicityMatt: Should be like this:

```suggestion
    cameraSerials: [...mastCams, ...driveCams, ...autoCams],
```

(and for the other page below)

src/ros/nova-gui/nova-gui/src/views/urc/URCAutonomousNavigationView.tsx
```diff
@@ -1,8 +1,23 @@
 import React from "react";
+import { Cartographer } from "../../components/maps/Cartographer/Cartographer";
+import AutoStatus from "../../components/maps/Cartographer/components/AutoStatus";
```
FelicityMatt: Could we move the auto components out of the cartographer folder.

I created an auto folder here:
https://github.com/MonashNovaRover/nova/tree/master/src/ros/nova-gui/nova-gui/src/components/auto

Perhaps move these components into a `cartographer` folder within auto? If you can think of a better name go ahead (`cartographer-components`?)

FelicityMatt (changes requested):
Overall looking really good, great job extracting everything out. I've left a couple of small comments, but the main thing is that the maps doesn't reach the bottom of the screen when the bottom overlay is down.

Screenshot From 2025-08-11 12-00-39

src/ros/nova-gui/nova-gui/src/components/maps/Cartographer/components/BottomOverlay.tsx
Matthewheekh: @FelicityMatt wouldn't there be too much movement if the map moves depending on the overlay?

src/ros/nova-gui/nova-gui/src/components/maps/Cartographer/components/CartographerGoalModal.tsx
Matthewheekh: @FelicityMatt yup, also I found an error with how we're creating the MapPoints, it uses the name as the key when displaying in table, but whenever we create a new point with no name, it defaults to a name based on the number of points we currently have + 1, so for example:

we create point1, point2, point3
then we delete point1
the next point will now by default have a name of point3 since we have 2 points previously
so now we have duplicated the point3 name

just wondering if we should fix this or do we always just specify a name when we create points and this scenario wont happen

FelicityMatt: oh interesting, I think we normally do specify names, but if you can't think of a way to fix this then go ahead

src/ros/nova-gui/nova-gui/src/components/maps/Cartographer/components/BottomOverlay.tsx
FelicityMatt: mm true, just worried we're gonna cut off the bottom of the map, but I guess it's ok?

FelicityMatt (commented):
Can we move `CartographerGoalModal.tsx` and `GoalPublisher.tsx` to the auto cartographer components folder and then can you rebase off of master!

Just tested and it looks good to me! happy to merge after the above two things have been done

FelicityMatt (approved):
lgtm!

PR #658, opened 2025-08-12
Complete drive overhaul

### LINKED ISSUE #587: Fix/rework teleop_drive_joy and drive controllers
Currently teleop_drive_joy and pivot_drive_controller is messy and broken. Time to fix it!
In the process, other drive controllers will also be reworked to accommodate the changes.

Development is happening on `feat/improved-drive`

Tags for Notion Integration:
tag:autonomous;tag:software;tag:rover;tag:control

tytia:
## Overview

It's time to retire old drive and switch to ros2_control controllers! **But why?**
- We currently use different drive controllers for manual and autonomous operation. The same controllers should be used to reduce repeated code.
- We should prefer ros2_control controllers as they are more standardised and will be easier to maintain in the future than our custom solution (old drive)

That being said, the previous ros2_control controller implementations we had for drive were **a huge mess** and **didn't work properly**. So that's what I've been working on for the past month!

Related issue: #587 

### [`drive/drive_controllers`](https://github.com/MonashNovaRover/nova/tree/feat/improved-drive/src/ros/rover/drive/drive_controllers)

Modified packages:
- `pivot_drive_controller`
- `diff_drive_controller`
- `strafe_drive_controller`

New packages:
- `nova_drive_controller_base`

### [`rover`](https://github.com/MonashNovaRover/nova/tree/feat/improved-drive/src/ros/rover)

Modified packages:
- `teleop_drive_joy`

New packages:
- `nova_controller_common`

### Summary of Changes

- Fixed/added speed, angular velocity and pivot position limiting to all controllers
    - This will prevent operators from breaking the ankles of the rover when trying to turn at a high speed
- Fixed speed calculations in `pivot_drive_controller` for small turning radii (previously the rover would drive at a much higher speed than intended)
- Moved all kinematics logic for `pivot_drive_controller` to a separate header, `kinematics.hpp`
- All controllers now inherit from a new base class, NovaDriveControllerBase. This makes the writing of new drive controllers in the future much MUCH easier.
- Created a new odometry class for all drive controllers to use, thereby individual controllers' odometry classes have been removed.
- DriveInput is no longer used, teleop_drive_joy now only sends Twist messages
- Other changes can be found in the documentation, i.e. READMEs and comments.

### Documentation

I've updated and added READMEs to every package that was created/modified. The comments within files will also provide context. Please take a look!

## Screenshots

https://github.com/user-attachments/assets/42ebfb41-abec-4095-831b-bbff7c766dec

## Steps to Test

1. Firstly, configure your controller following the steps in the `teleop_drive_joy` [README.md](https://github.com/MonashNovaRover/nova/tree/feat/improved-drive/src/ros/rover/drive/teleop_drive_joy#readme), unless you are using a Nintendo Switch Pro Controller, which is what I've been developing with and have configured already.
2. Run `ros2 launch teleop.launch.py config:=your_config`
    - you may have to specify `device_id:=1` if it loads your keyboard instead
    - 'your_config' should be the name of your config file, e.g. if your config file is `nintendo.config.yaml`, use `config:=nintendo`
3. Run `ros2 launch auto_bringup gazebo.launch.py camera:=false`
4. Happy driving!

### Notes on testing
- There's an issue with `diff_drive_controller`, but I suspect it may be due to the sim. Inspecting the wheel transforms via rviz, it seems as though incorrect speeds are being sent to the wheels but I've looked over the code many times and couldn't find anything that could cause such an issue.
- Please fiddle with parameters such as:
    - limiter params, e.g. speed, angular, pivot
    - `pivot_rate_tolerance` for `pivot_drive_controller`
    - any other params you're curious about
- Parameters are to be configured in [`auto/auto_bringup/params/controller.yaml`](https://github.com/MonashNovaRover/nova/blob/feat/improved-drive/src/ros/rover/auto/auto_bringup/params/controllers.yaml)
- A list of configurable parameters can be found in [nova_drive_controller_base.yaml](https://github.com/MonashNovaRover/nova/blob/feat/improved-drive/src/ros/rover/drive/drive_controllers/nova_drive_controller_base/src/nova_drive_controller_base.yaml). Configurable parameters specific to particular controllers may be found in controller_name.yaml. Currently only pivot_drive_controller specifies additional parameters in [pivot_drive_controller.yaml](https://github.com/MonashNovaRover/nova/blob/feat/improved-drive/src/ros/rover/drive/drive_controllers/pivot_drive_controller/src/pivot_drive_controller.yaml).

And observe the effects, verifying that everything is behaving as expected!

## TODO
- Test IRL on the rover
- Tune parameters for manual operation and create a separate .yaml file for manual operation
- Rewrite `teleop_drive_joy` to use `teleop_modular`
- Publish `DriveInfo` from `teleop_drive_joy`
- Change `base` and `drive` aliases/launch files to use new drive

## Memes

image

Tags for Notion Integration:
tag:autonomous;tag:software;tag:rover;tag:control


vic-bart:
What's the reasoning behind different controller layouts for wireless VS wired xbox controllers?

tytia:
@vic-bart The wireless xbox controller has not been updated; when I was testing on campus I couldn't find the wireless xbox controller so I configured and used the wired one. I made a separate controller layout for the wired one in case the inputs are slightly different

BaileyChessum:
Can't test with my ps5 controller, as it has the dpad (needed for speed change) as buttons, but teleop_drive_joy used axes. `teleop_modular` please please please please please plase please plaease please. It has axis to button remapping and button to axis remapping and `increment_axis` commands

BaileyChessum:
I tried making a very simple `teleop_drive` on `feat/teleop-drive`. A bit jank.
Works for all controllers.

BaileyChessum:
Setting up my control scheme like:

```yaml
linear:
  x: LEFTY
  y: LEFTX
  z: 0.0
angular:
  x: 0.0
  y: 0.0
  z: RIGHTY
```

and setting pivot_drive_controller's `autonomous_mode` param to false seems to produce behavior similar to what is expected, but not quite right.

Holding RIGHTY=1 doesnt make the pivots go to a full circle, and seems to make the rover turn left when i move the joystick right (but it goes the correct direction when using diff drive controller).

Not sure what the correct values to provide to it are? 

setting `autonomous_mode` to true brings it closer to expected behaviour, but introduces unexpected pauses in drive when pivoting, and sometimes the pivots have a moment and vibrate strangely. Pivoting to the right, holding forward, then holding backwards will also cause the pivots to unnecessarily do a 180 degree turn.

vic-bart:
@tytia This PR is now blocking [Moving the rover with ROS2 Control](https://github.com/MonashNovaRover/unity-sim/issues/5).

vic-bart:
Met with Terry today: this PR is pending testing on rover, which is currently being blocked by getting BLCMDs to work (which JZ says should be ready within the week hopefully :crossed_fingers:)

tytia:
New drive has been tested on the rover IRL and works. It is safe to merge as it doesn't affect old drive. I will make a new branch for developing the new launch files and parameters to streamline the process of launching new drive.

vic-bart:
@tytia comments from my initial PR review:
- Renaming nova_diff_drive_controller to diff_drive_controller. Makes sense, but consider (if you haven't) that this name was chosen to avoid conflicting with ros2_control's existing diff_drive_controller.
- Why is this modifying teleop_drive_joy? I'm probably just misunderstanding the function of teleop_modular, because I was under the belief that teleop_modular replaces teleop_drive_joy.

Otherwise, lgtm, merge! And it has documentation, even better!

tytia:
> @tytia comments from my initial PR review:
> 
> * Renaming nova_diff_drive_controller to diff_drive_controller. Makes sense, but consider (if you haven't) that this name was chosen to avoid conflicting with ros2_control's existing diff_drive_controller.
> * Why is this modifying teleop_drive_joy? I'm probably just misunderstanding the function of teleop_modular, because I was under the belief that teleop_modular replaces teleop_drive_joy.
> 
> Otherwise, lgtm, merge! And it has documentation, even better!

- Yep I've considered the name conflict with ros2_control's diff drive controller, but I decided to make the change anyway because in no situation would we use ros2_control's diff drive controller over our own one.
- So teleop_modular has actually not been implemented for drive yet, due to some missing features we need to implement in teleop_modular. So for the time being, drive will continue to use teleop_drive_joy.

PR #659, opened 2025-08-12
New arm launch file configuration

BaileyChessum:
## Description

This PR adds the necessary launch and parameter files to be able to run the new arm in sim.

## Screenshots

[Screencast From 2025-08-13 00-27-53.webm](https://github.com/user-attachments/assets/8c7d7fc9-c7ed-4174-aa4d-49159b205cc0)

## Steps to Test

In one terminal, run:
```sh
```

In another terminal, run:
```sh
mros2 launch teleop_arm teleop.launch.py teleop_params:=/home/nova/nova/src/ros/rover/arm/teleop_arm/params/teleop.yaml log_inputs:=True
```

Then plug in an xbox or ps5 controller. 

You can find the control scheme under `src/ros/rover/arm/teleop_arm/params/teleop.yaml`.

### DIFF
--- a/src/ros/rover/arm/arm_bringup/launch/control.launch.py
+++ b/src/ros/rover/arm/arm_bringup/launch/control.launch.py
@@ -35,6 +35,13 @@ def launch_setup(context, *args, **kwargs):
     use_mock_hardware = LaunchConfiguration('use_mock_hardware')
     robot_name = LaunchConfiguration('robot_name')
 
+    show_colours_additional_env = {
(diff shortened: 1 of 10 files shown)


Redecorating:
when i use the left trigger to enter ik mode the target pose jumps around. Otherwise this seems to work with an xbox controller although I haven't been able to get the thrustmasters to do anything

BaileyChessum:
Included with #669

Issue #657, opened 2025-08-11
Open source the repo


tytia:
### **Context**

We want to open source the repo to increase visibility and get some GitHub benefits (such as mutliple assignees per issue).
See [Notion Discussion Point](https://www.notion.so/Open-Sourcing-231b713961718023a11dde90efc24225).

### **TODO**

- Get consent from past contributors
- Decide on a license
- Remove secrets (maybe a separate private repo for secrets and call it as a submodule?)

Tags for Notion Integration:
tag:autonomous;tag:software

vic-bart:
Made a start on getting consent from past contributers. I cloned a bunch of repos and used `git log --format='%an,%ae' | sort -u` to find all the authors (note that some contributers were not identified as such, not sure why 🤷 but it means that we might miss people?).

Anyway here's what I found:
- [x] nova
	- GitHub:
		- Anthony Lew, ant87518@gmail.com
		- Bailey Chessum, bailey.chessum1@gmail.co
		- Connor Macdougall, cmac0046@student.monash.edu
		- Felicity Matthews, felicitymatthews30@gmail.com
		- Joshua Leivenzon, jlei0014@student.monash.edu
		- Jared Landau, jlan0046@student.monash.edu
		- Jonathan Chin, jchi0097@student.monash.edu
		- Kabilan Velmurugan Sujatha, kabiniki4849@gmail.com
		- Kuhu Tosniwal, ktos0001@student.monash.edu
		- Matthew Hee, mhee0019@student.monash.edu
		- Orlando Chamberlain, ocha0025@student.monash.edu
		- Rohit Pilakkat, rohit.pilakkat@gmail.com
		- Brandon Chung, bchu0028@student.monash.edu
		- Tash Lee, tlee0160@student.monash.edu
		- Terry Tian, ttia0006@student.monash.edu
		- Victor Bartlinski, vicbart08@gmail.com
	- Undocumented:
		- Chetan Edupalli, Cedu0001@student.monash.edu
		- Tarik Thomas, ttho0026@student.monash.edu
		- Abby Ahmed, aahm0045@student.monash.edu
		- …
- [x] rover
	- GitHub:
		- Abigail Rose, abigail.rose@me.com
		- Alexander Li, li420087@gmail.com
		- Unknown, apri0009@student.monash.edu
		- Ariel Elbaum, ariel.elbaum@outlook.com
		- Unknown (BAMS?), modulepackdata@energyrenaissance.com
[...]

PR #668, opened 2025-08-19
Hiding passwords

vic-bart:
## Description

Issue #657 

I've created a separate private repo called "secrets" and added it as a submodule to nova (found at nova/secrets) and moved every secret we know of to this repo, substituting their original code with code that reads from file-relative filepaths pointing towards the corresponding secret in the repo.

## Steps to Test

Not really sure how to test this entirely :sweat_smile: so far I've been making the change, logging the password, and checking it's the same. I've done this for every change where it's possible (e.g. reolink-ctl, radio_monitor.py) but not for the ones affecting hydra.

### DIFF
--- a/nixfiles/ci/jobsets/docker.nix
+++ b/nixfiles/ci/jobsets/docker.nix
@@ -32,7 +32,8 @@ let
           installer.cloneConfig = false;
 
           nova = {
-            substituters.nova.password = "tFH6J!#HhrYc3&^m";
+            substituters.nova.password = builtins.readFile ../../../secrets/hydra-password.txt;
+            
             desktop.enable = false;
           };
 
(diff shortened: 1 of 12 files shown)


nixfiles/ci/jobsets/docker.nix
```diff
@@ -32,7 +32,8 @@ let
           installer.cloneConfig = false;
 
           nova = {
-            substituters.nova.password = "tFH6J!#HhrYc3&^m";
+            substituters.nova.password = builtins.readFile ../../../secrets/hydra-password.txt;
```
FelicityMatt, line 35: have you been able to test any of these nix secret references? I'm assuming if one works they should all work. And I mean hydra is already broken rn so what's one more thing...

FelicityMatt (commented):
lgtm, but yeah not sure how to test hydra or if we can so may as well test after merging?

FelicityMatt:
@vic-bart hows this going? if you rebase can we now test it works with hydra?

FelicityMatt:
@vic-bart what's the current status of this?

vic-bart:
Sorry! Yes, this works with hydra:
image

PR #669, opened 2025-08-20
fix: nova_twistmapper 

BaileyChessum:
## Description

Merges changes to make twist.angular be applied via an angle-axis relative to some reference frame_id in nova_twistmapper

## Steps to Test

In one terminal, run:
```sh
```

In another terminal, run:
```sh
mros2 launch teleop_arm teleop.launch.py teleop_params:=/home/nova/nova/src/ros/rover/arm/teleop_arm/params/teleop.yaml log_inputs:=True
```

Then plug in an xbox or ps5 controller. 

You can find the control scheme under `src/ros/rover/arm/teleop_arm/params/teleop.yaml`.

## Memes

Tags for Notion Integration:
tag:arm;tag:control


Redecorating (approved):
this fixed it

PR #671, opened 2025-08-20
feat: local rover_description in URDF via rover_description_dir xacro arg

BaileyChessum:
## Description

This PR is intended to make it easier to modify configuration files (.urdf.xacro, .yaml param files) on the fly without needing to rebuild the workspace.

In launch files for the arm, simply use the param `local:=true` to use local files, rather than files from the nix store

## Steps to Test

try run 

```
ros2 launch teleop_arm teleop.launch.py local:=True
```

then change `teleop_arm/params/teleop.yaml` to have invalid syntax and run it again. The invalid syntax should cause it to fail.

Then, do the same with the rover's URDF:
```
ros2 launch arm_bringup control.launch.py rviz:=True use_mock_hardware:=True local:=True
```
Invalid syntax should cause it to fail.

## Memes

Tags for Notion Integration:
tag:arm;tag:rover;tag:ready_for_review

### DIFF
--- a/src/ros/rover/arm/arm_bringup/launch/control.launch.py
+++ b/src/ros/rover/arm/arm_bringup/launch/control.launch.py
@@ -44,6 +44,12 @@ def launch_setup(context, *args, **kwargs):
         'RCUTILS_CONSOLE_OUTPUT_FORMAT': '[{severity}] [{name}] {message}',
     }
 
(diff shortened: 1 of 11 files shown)


Redecorating (approved):
u should `sudo mount -o remount,rw /nix` and then just edit the urdf in the nix store :)

PR #672, opened 2025-08-26
feat: Python Control 2

### LINKED ISSUE #664: Design and Implement Python Control v2
### **TODO**

-

BaileyChessum:
## Description

Implements a second iteration of python control, using a model similar to that of ros2_control.

## Steps to Test

```
ros2 run science control_test.py
```


src/ros/rover/nova_generic/python_control2/python_control2/controllers/Controller.py
```diff
@@ -0,0 +1,80 @@
...
+        """ Overrides construction of Controller instances to defer calling __init__ until contexts are available. """
+        return DeferredConstructor(cls, *args, **kwargs)
+
+    def __init__(self, contexts: Contexts):
+        """ Constructor, deferred until the control manager has been spun.
+        If you override this method, and want to add your own arguments, just make sure contexts is the FIRST arg
+
+        :param contexts: A collection of dependency injection class instances you can index by class type.
+        """
+        print("Controller base init")
```
FelicityMatt: do we need this?

src/ros/rover/nova_generic/python_control2/python_control2/controllers/DeferredConstructor.py
```diff
@@ -0,0 +1,64 @@
...
+        provided.
+        :param name: The name of the constructed class instance. Uses self.name if None is given, and generates a name
+        from the class name if self.name is also None
+        :return: The instance of the class
+        """
+        # Ensure name and node have values
+        if name is None:
+            name = self.name
+        if name is None:
+            name = camel_to_snake(self.cls.__name__.lower())
```
FelicityMatt, line 36: why are we doing this? I feel like that could really confuse people if they name something one way and then it appears differently? I do understand convention but I'm not sure we should enforce it like this?

FelicityMatt (commented):
two comments otherwise lgtm

src/ros/rover/nova_generic/python_control2/python_control2/controller_manager/Interface.py
```diff
@@ -0,0 +1,44 @@
...
+
+class InterfaceCollection:
+    """ Wrapper for a dictionary of Interfaces. Represents Command and State Interfaces. """
+
+    def __init__(self, default_value: T=0):
+        self._values: dict[str, Interface] = {}
+        self.default_value = default_value
+
+        # Event called whenever we retrieve an interface. For internal use only.
+        self.on_get_item: Event[[Interface]] = Event()
```
TJ-Lee-M, line 26: Does retrieving an interface happen when the controller in spun? Or when hardware is connected? (Im trying to understand when we set_populated in the controller manager)

TJ-Lee-M: Answer: 
on_get_item() is triggered when data is received by the controller (from a hardware interface) or visaversa. So calling set_populated when this happens is to track that the controller and interface have successfully exchanged data

src/ros/rover/nova_generic/python_control2/python_control2/controllers/Controller.py
BaileyChessum: no. whoops

src/ros/rover/nova_generic/python_control2/python_control2/controllers/DeferredConstructor.py
BaileyChessum: Only happens when the user uses the deferred constructor directly. If you're using this stuff directly, you already know what's going on enough to not get confused by this. The controller manager builder will always ask for a name to use, and then use that name. This will never happen when using the controller manager builder

FelicityMatt (approved):
lgtm! Lets try it out and fix what breaks!

Only thing is that do we want to commit the science/control_test.py to master?

PR #684, opened 2025-09-06
ci lib: reenable aliases for release-lib

Redecorating:
https://github.com/NixOS/nixpkgs/commit/2116eb760b44f4e44d500f32ab9f2570fc586dac

I think this is what broke it, if aliases are disabled then the hostPlatform etc attributes don't exist.


Redecorating:
idk how to test without merging so i merge now good luck everyone

PR #697, opened 2025-09-14
ci: slave: configure builduser decleratively

Redecorating:
## Description

wip moving the stuff I did to make hydra have an x86 build worker into nixfiles

~~need to make it so you can configure nova.ci.master.slaves = {something}~~
~~need to make the ssh keys for the hydra slave not hard coded~~

~~need to do actual port forwarding to get rid of the stupid nat punching script~~ port forwardning didn't work consistently...

## Memes

Tags for Notion Integration:
tag:software;tag:nixfiles;tag:ready_for_review


Redecorating:
See also: https://github.com/MonashNovaRover/nix-configs/tree/hydra, https://github.com/MonashNovaRover/nix-configs/tree/workshop-build-pc

Redecorating:
it sounds like the execs get emails from oracle so they might be able to do port forwarding and then the messy bits in this can be removed

FelicityMatt:
@Redecorating what's the status of this? is this still needed?

Redecorating:
i think i want to move the wireguard config stuff to nix-confgs because its messy and then we can merge just the main build slave stuffOn 2 Feb 2026, at 10:21 am, Felicity Matthews ***@***.***> wrote:﻿FelicityMatt left a comment (MonashNovaRover/nova#697)
@Redecorating what's the status of this? is this still needed?

PR #704, opened 2025-09-20
feat: allow local teleop_modular source directory to be specified

FelicityMatt:
## Description

You can set the `TELEOP_PATH` env variable to point to your local source code for teleop_modular to make ws-build build with that teleop_modular source:

```sh
export TELEOP_PATH:=~/path/to/teleop_modular
```


BaileyChessum (approved):
I might be biased since I made the changes, but LGTM

PR #712, opened 2025-09-22
Feat/can script canable

Redecorating:
## Description
make the can script support canable
also make a `can mv [src] [dst]` for renaming the can buses.

### DIFF
--- a/nixfiles/modules/home/macros/can.sh
+++ b/nixfiles/modules/home/macros/can.sh
@@ -124,18 +124,19 @@ then
 
 elif [[ $failed != "1" ]]
 then
-
-    # Run the CAN
-    sudo modprobe can
-    sudo modprobe can-raw
-    sudo modprobe mttcan
-
     # If turning on CAN
     if [[ $command == "start" ]]
     then
+        DRIVER=$(basename $(readlink "/sys/class/net/$can/device/driver"))
+        # canable does not support berr-reporting
+        if [ A$DRIVER = Ags_usb ]; then
+          EXTRA_OPTIONS=""
+        else
+          EXTRA_OPTIONS="berr-reporting on"
+        fi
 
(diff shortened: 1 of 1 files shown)


nixfiles/modules/home/macros/can.sh
```diff
@@ -124,18 +155,19 @@ then
 
 elif [[ $failed != "1" ]]
 then
-
-    # Run the CAN
-    sudo modprobe can
```
FelicityMatt, line 129: what did these lines do?

nixfiles/modules/home/macros/can.sh
```diff
@@ -6,9 +6,15 @@
...
 #
+# You can also use mv to rename any can bus. It
+# will retain an alias to its original name +
+# "_orig" eg can0_orig which you can see in
+# the output of `ip link`.
+#
 # Some examples could be:
 #   can start can0
 #   can stop vcan0
+#   can mv can0 can1 (rename can0 to can1)
```
FelicityMatt, line 17: this would be amazing thank you!

FelicityMatt (commented):
lgtm, but have you been able to test it works with both cannable and normally?

Redecorating:
tested on cannable, need to test on orin/jetson

nixfiles/modules/home/macros/can.sh
Redecorating: manually load the kernel modules, which would already have been done when the kernel detects the can devices. In particular, this bit errors on platforms other than orin/tx2 as they don't have the mttcan driver which the code originally assumed all of our can devices would use.

Redecorating:
tested this on the orin. ive also noticed that you can't rename can busses when they are already started or it fails. make sure we can stop then can mv then can start

PR #713, opened 2025-09-23
removing guiEnabled

vic-bart:


vic-bart:
Tested by Chetan, works

PR #714, opened 2025-09-24
New drive launch files and aliases

tytia:
## Description

Added launch files and aliases for new drive. They have been tested and they work. Some existing aliases launch files have been modified to ensure things still work. Please ensure I didn't unintentionally make any breaking changes elsewhere 😱


nixfiles/modules/home/macros/default.nix
```diff
@@ -167,12 +169,11 @@ in
           launch-sim = "~/Builds/master/bin/ros2 launch auto_bringup everything.launch.py";
           launch-auto-hardware = "~/Builds/master/bin/ros2 launch auto_bringup hardware.launch.py";
           launch-auto-software = "~/Builds/master/bin/ros2 launch auto_bringup software.launch.py";
-          launch-control = "~/Builds/master/bin/ros2 launch auto_bringup control.launch.py";
```
TJ-Lee-M, line 170: autos version of old drive
launch-auto-drive is the same thing

src/ros/rover/drive/drive_bringup/README.md
```diff
@@ -1 +1,4 @@
-TODO
+# New Drive
```
Redecorating: maybe dont call it new drive because it will not always be new

Redecorating: i say this as chassis keeps saying new drive and i dont always know if its the same new thing or a different new thing to their last new drive that's "almost done"

TJ-Lee-M: good point- what about "teleop drive"?

TJ-Lee-M (approved):
I second making a more desciptive/specific name than new drive
Otherwise lgtm :saluting_face:

src/ros/rover/drive/drive_bringup/README.md
tytia: changed to "Drive Bringup", realised the only reason it was named new drive was because I copied arm_bringup and the README for arm_bringup says "New Armware"

PR #716, opened 2025-09-25
nixfiles: jobsets: isos: resolve name conflict

Redecorating:
make the iso name a lib.mkForce

error: The option `image.baseName' has conflicting definition values:
- In `/nix/store/081zspnjiwbq85jqhx89gkk2nbyrfg3y-source/nixos/modules/installer/cd-dvd/iso-image.nix': "nixos-graphical-workspace"
- In `/nix/store/081zspnjiwbq85jqhx89gkk2nbyrfg3y-source/nixos/modules/installer/cd-dvd/iso-image.nix': "nixos-nova-nova-25.11pre-git-x86_64-linux" Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.

### DIFF
--- a/nixfiles/ci/jobsets/isos.nix
+++ b/nixfiles/ci/jobsets/isos.nix
@@ -41,7 +41,7 @@ let
               ];
 
               isoImage = {
-                isoBaseName = "nixos${lib.optionalString graphical "-graphical"}${lib.optionalString includeWorkspace "-workspace"}";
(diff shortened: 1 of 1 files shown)


Redecorating:
hydra doesn't build this pr...

I think `nixfiles/ci/default.nix` has `planRosDistroAndPrJobsets` only used for misc and workspaces, so the isos jobset doesn't get built for PRs. @hacker1024 do you know if that's intentional? I'm wanting to get hydra to build the jetpack 6 iso for devkit but for now im building it manually on the build slave pc instead.

PR #718, opened 2025-09-26
Centralise all aliases to macros default nix

V01DBREAKER:
We had aliases defined in two different places with some overlap/duplicates.
This removes them from rover's default.nix and keeps it all in nixfiles/home/macro/default.nix

## Memes
image

Tags for Notion Integration:
tag:nixfiles;tag:ready_for_review

### DIFF
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -46,28 +46,16 @@ in
           nova = "cd ${cfg.sourceDir}/..";
           nixfiles = "cd ${cfg.nixfileDir}";
           rover = "${nova}/src/ros/rover";
-          core = "${nova}/src/ros/rover/core";
-          control = "${nova}/src/ros/rover/control";
-          electronics = "${nova}/src/ros/rover/electronics";
-          elec = electronics;
-          visualisation = "${nova}/src/ros/visualisation";
-          visualization = visualisation;
-          vis = visualisation;
-          viz = visualisation;
+          arm = "${nova}/src/ros/rover/arm";
+          autonomous = "${nova}/src/ros/rover/auto";
+          auto = autonomous;
+          chassis = "${nova}/src/ros/rover/chassis";
           science = "${nova}/src/ros/rover/science";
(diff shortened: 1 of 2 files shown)


vic-bart:
Note that I'm busy right now, but please don't merge before I've made some edits

Redecorating (approved):

vic-bart:
Removed deprecated (old and no longer used) aliases.

V01DBREAKER:
I removed some more old aliases and stubs and made a couple new ones too, please review again @vic-bart @Redecorating

nixfiles/modules/home/macros/default.nix
```diff
@@ -40,32 +40,20 @@ in
...
           rover = "${nova}/src/ros/rover";
-          core = "${nova}/src/ros/rover/core";
-          control = "${nova}/src/ros/rover/control";
-          electronics = "${nova}/src/ros/rover/electronics";
-          elec = electronics;
-          visualisation = "${nova}/src/ros/visualisation";
-          visualization = visualisation;
-          vis = visualisation;
-          viz = visualisation;
+          arm = "${nova}/src/ros/rover/arm";
```
Redecorating, line 49: do people actually use these? if they have been broken for a while without complaint then maybe we don't need them

Redecorating (approved):

nixfiles/modules/home/macros/default.nix
V01DBREAKER: So true, but i assume they're for power users

vic-bart:
Lgtm, probably better to leave ilmenite-shell as predict-shell but it should be fine.

TJ-Lee-M:
I agree- looks good but I dont like ilmenite-shell.
I think predict-shell is good but if you really want to make it more descriptive, what about predict-shell-science, or predict-shell-arc ? longer but easier to see typos/spelling errors

V01DBREAKER:
LGTM (Lets gamble the moon). also i changed ilmenite shell back to predict shell

V01DBREAKER:
image
Should we remove these before open source? @TJ-Lee-M

PR #719, opened 2025-09-26
Feat/rtabmap launch

Chetan-Edupalli:
## Description

everything.launch.py now runs rtabmap :partying_face: 

This has been a know issue that rtabmap must be run before other stack runs
## Steps to Test

We now run `ros2 launch auto_bringup everything.launch.py rtabmap:=true` to enable rtabmap (this disables gps). 
everything.launch.py checks for topics published by rtabmap before starting other launch files in the stack.

To run using gps mode, just remove the rtabmap argument from the command, so that would be `ros2 launch auto_bringup everything.launch.py`

## Memes
image

Tags for Notion Integration:
tag:autonomous;tag:ready_for_review

### DIFF
--- a/src/ros/rover/auto/auto_bringup/CMakeLists.txt
+++ b/src/ros/rover/auto/auto_bringup/CMakeLists.txt
@@ -28,6 +28,7 @@ install(DIRECTORY
   resources
   rviz
   maps
+  topic
   DESTINATION share/${PROJECT_NAME}
 )
 
--- a/src/ros/rover/auto/auto_bringup/launch/everything.launch.py
+++ b/src/ros/rover/auto/auto_bringup/launch/everything.launch.py
@@ -14,7 +14,7 @@ CREATION:	27/04/2023
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 '''
 from launch import LaunchDescription
-from launch.actions import DeclareLaunchArgument, IncludeLaunchDescription, OpaqueFunction
+from launch.actions import DeclareLaunchArgument, IncludeLaunchDescription, OpaqueFunction, GroupAction, TimerAction
 from launch.substitutions import  PathJoinSubstitution, LaunchConfiguration
 from launch.conditions import IfCondition, UnlessCondition
 from launch.launch_description_sources import PythonLaunchDescriptionSource
@@ -41,8 +41,9 @@ def launch_setup(context, *args, **kwargs):
     sim_params = LaunchConfiguration('sim_params')
     use_respawn = LaunchConfiguration('use_respawn')
     world = LaunchConfiguration('world')
+    rtabmap = LaunchConfiguration('rtabmap')
 
-    return [
+    auto_bringup_common_nodes = GroupAction([
         IncludeLaunchDescription(
             condition=IfCondition(gazebo),
             launch_description_source=PythonLaunchDescriptionSource(PathJoinSubstitution([auto_bringup_dir, 'launch', 'gazebo.launch.py'])),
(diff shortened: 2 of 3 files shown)


vic-bart:
@Chetan-Edupalli I don't understand how this code could have run on your machine? Because running `ros2 launch auto_bringup everything.launch.py rtabmap:=True` shows wait_for_topic.py failing:

```
[wait_for_cameras_topic-2] Traceback (most recent call last):
[wait_for_cameras_topic-2]   File "/nix/store/fjwxpgmiks5m9g50sxiinv5i5lsz0b5r-ros-jazzy-nova-workspace/share/auto_bringup/topic/wait_for_topic.py", line 86, in <module>
[wait_for_cameras_topic-2]     main()
[wait_for_cameras_topic-2]   File "/nix/store/fjwxpgmiks5m9g50sxiinv5i5lsz0b5r-ros-jazzy-nova-workspace/share/auto_bringup/topic/wait_for_topic.py", line 77, in main
[wait_for_cameras_topic-2]     success = node.spin_until_ready()
[wait_for_cameras_topic-2]               ^^^^^^^^^^^^^^^^^^^^^^^
[wait_for_cameras_topic-2]   File "/nix/store/fjwxpgmiks5m9g50sxiinv5i5lsz0b5r-ros-jazzy-nova-workspace/share/auto_bringup/topic/wait_for_topic.py", line 64, in spin_until_ready
[wait_for_cameras_topic-2]     if self.all_topics_active():
[wait_for_cameras_topic-2]        ^^^^^^^^^^^^^^^^^^^^^^^^
[wait_for_cameras_topic-2]   File "/nix/store/fjwxpgmiks5m9g50sxiinv5i5lsz0b5r-ros-jazzy-nova-workspace/share/auto_bringup/topic/wait_for_topic.py", line 58, in all_topics_active
[wait_for_cameras_topic-2]     available_topics = [t.name for t in self.get_topic_names_and_types()]
[wait_for_cameras_topic-2]                         ^^^^^^
[wait_for_cameras_topic-2] AttributeError: 'tuple' object has no attribute 'name'
```

And when I look for the root cause, it's because [`get_topic_names_and_types()`](https://github.com/ros2/rclpy/blob/10f70c9674933159b348524d8872d4bc3d542116/rclpy/rclpy/node.py#L2146) returns a list of tuples `[(topic_name, [topic_types]), ...]`:

```py
def get_topic_names_and_types(self, no_demangle: bool = False) -> List[Tuple[str, List[str]]]:
    """
    Get a list of discovered topic names and types.

    :param no_demangle: If ``True``, then topic names and types returned will not be demangled.
    :return: List of tuples.
      The first element of each tuple is the topic name and the second element is a list of
      topic types.
    """
    with self.handle:
        return _rclpy.rclpy_get_topic_names_and_types(self.handle, no_demangle)
```

And the code is trying to access an attribute with the key "name" (which would never work). So this code:

```py
available_topics = [t.name for t in self.get_topic_names_and_types()]
```

Should actually be this code:

```py
available_topics = [t[0] for t in self.get_topic_names_and_types()]
```

In order to work. I thought maybe you were using an older version of rclpy, but the git blame shows the return type last changed 6 years ago, so I genuinely can't figure out what happened :sweat_smile: Can you bring your laptop and walk me through how it worked or is it on your home PC?

[EDIT] We have no idea how it worked earlier :skull:

vic-bart:
@Chetan-Edupalli @KABILAN235 can I get some clarification on the race condition being resolved here? 
I was under the belief that the issue is rtabmap was launching before the cameras launched, causing it to fail. But examing the changes here indicate the issue is actually the opposite: failure was caused by the cameras launching before rtabmap?

[EDIT] Yes, failure was caused by the cameras launching before rtabmap.

vic-bart:
Cleaned up the launch file a bit, built, tested, lgtm :+1:

PR #720, opened 2025-09-27
Tileserver fix for cartographer with nova-gui readme and alias adjustments

V01DBREAKER:
## Description
This PR fixes the cartographer tool in nova-gui by using tileserver-gl-light instead of tileserver-gl.
Tileserver-gl was broken due to node 18 EOL and other dependency hell schenanigans.

As a result, the nova-gui README and aliases were updated to match the new process.

Also, the `gui` alias was updated to actually match the readme's definition of it.

## Changelogs
Added new tileserver-gl-light and tileserver-gl-styles to the yarn lock file
Updated the gui aliases
Updated the gui readme

## Screenshots
image

## Steps to Test
Follow the updated readme for the nova-gui in this PR.
Make sure to `sudo nixos-rebuild switch` to test the aliases!

## Memes
node-modules-app-performance_

Tags for Notion Integration:

### DIFF
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -56,7 +56,7 @@ in
           camerasdir = "${nova}/src/ros/cameras2";
           autonomous = "${nova}/src/ros/rover/autonomous";
           auto = autonomous;
-          gui = "${nova}/src/ros/nova-gui";
+          gui = "${nova}/src/ros/nova-gui/nova-gui";
           tutorials = "${nova}/src/ros/tutorials";
           pic = "${nova}/src/other/pics";
           pics = pic;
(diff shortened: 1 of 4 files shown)


nixfiles/modules/home/macros/default.nix
```diff
@@ -135,10 +135,8 @@ in
           gui-run = "yarn --cwd ~/nova/src/ros/nova-gui/nova-gui dev";
```
FelicityMatt, line 135: while we're here can you add an alias 'gui-yarn' that is this alias minus the dev? it should install all the dependencies and allow us to do any other yarn commands from anywhere

FelicityMatt (commented):
i'll try give it a try on my laptop soon but preliminarily looks good, i might rework the readme at some point to be a bit clearer.

FelicityMatt (approved):
Tested it and it LGTM.

I added a `gui-yarn` alias and updated the `gui-tilerun` alias to include the --file argument, feel free to revert the changes if don't want that.
https://github.com/MonashNovaRover/nova/pull/720/commits/dc50d8547797e65f2faa1f92e25704db461d3524

V01DBREAKER:
I like that change, probably means it'll fail if you don't put a file path in which is good

PR #727, opened 2025-10-07
test: Taipan hardware testing changes

BaileyChessum:
## Description

This branch is for the testing of Taipan on Wednesday October 8th.

## Memes

Tags for Notion Integration:
tag:arm;tag:control

### DIFF
--- a/src/ros/rover/arm/arm_bringup/params/new.controllers.yaml
+++ b/src/ros/rover/arm/arm_bringup/params/new.controllers.yaml
@@ -31,7 +31,7 @@ arm:
         - J5
         - J6
       use_position_control: false
-      use_collision_limits: true
+      use_collision_limits: false
       use_limits: true
 
   nova_arm_position_controller:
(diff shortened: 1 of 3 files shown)


src/ros/rover/arm/arm_bringup/params/new.controllers.yaml
```diff
@@ -31,7 +31,7 @@ arm:
         - J5
         - J6
       use_position_control: false
-      use_collision_limits: true
+      use_collision_limits: false
```
FelicityMatt, line 34: do we want to keep this off in master?

FelicityMatt (commented):
one question, otherwise lgtm

src/ros/rover/arm/arm_bringup/params/new.controllers.yaml
BaileyChessum: they dont have abcoders on the arm.

PR #730, opened 2025-10-10
hotfix until electrical increases pivot limits

tytia:
## Description

There is currently a miniscule (think 8th or 10th decimal point) inaccuracy between the pivot offset defined in the hardware interfaces and the actual pivot offset IRL, which causes the pivots IRL to flip 180 degrees back to stay within limits when sent a full left/right turn (turning radius = 0).

The calculations in the drive code should actually be correct, this is a temporary hotfix until electrical increases the pivot limits. When that happens, this line should be removed.


BaileyChessum:
lgmt

PR #733, opened 2025-10-11
Fix new drive

tytia:
## Description

New drive had an issue where pivots were really jittery. @BaileyChessum found out the problem was that the values in the previous value queues were being passed in the wrong order to the limiters, causing unwanted oscillating behaviour.

This is now fixed! Fixed and tested during the auto/lidar field test on 11/10/25.


tytia (commented):
lgtm

PR #748, opened 2025-10-24
feat: switch to game_controller_node and implement handbrake

### LINKED ISSUE #747: Use game_controller_node instead of joy_node
### **Context**

Switching to game controller node means that custom configs for each game pad are no longer necessary as this node uses SDL's built-in mapping database instead.

### **TODO**

- change corresponding node in launch file
- remove custom configs for game pads
- change the default mapping used by teleop_drive_joy to match the structure of messages published by game controller node

Tags for Notion Integration:
tag:software;tag:rover

Glass2721:
## Description

Switch from using `joy_node` to `game_controller_node` in `teleop_drive_joy` (so configs for each controller are no longer necessary), and implemented handbrake. Relevant configs, launch files and build files have been changed or removed.

## Steps to Test

With any game pad connected:

`ros2 launch auto_bringup gazebo.launch.py camera:=false`

`ros2 launch teleop_drive_joy teleop.launch.py`

Controls should work as expected, including right trigger (handbrake) which should decrease linear velocity by 40%.

## Memes

67

Tags for Notion Integration:
tag:software;tag:rover;tag:ready_for_review


src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.cpp
```diff
@@ -182,28 +182,41 @@ void TeleopDriveJoy::map_button_callbacks()
   {
     switch_controller(DriveMode::DIFF);
   };
-  // mark axis buttons as negative
-  button_callbacks_[-(params_.axis_speed_change_fine)] =
-    [this](const sensor_msgs::msg::Joy::SharedPtr joy_msg)
+
+  auto changeSpeed = [this](double speed_change)
```
tytia: Keep naming conventions consistent, i.e. rename to `change_speed`

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.cpp
```diff
@@ -228,7 +241,7 @@ void TeleopDriveJoy::handle_button_callbacks(const sensor_msgs::msg::Joy::Shared
   auto isPressedAndDebounced = [this, &now, joy_msg](int button_index)
   {
     bool pressed = (button_index >= 0 && joy_msg->buttons[button_index]) ||
-                   (button_index < 0 && joy_msg->axes[std::abs(button_index)]);
+                   (button_index < 0 && joy_msg->axes[std::abs(button_index)] < -params_.trigger_pressed_threshold);
```
tytia: I think using `std::abs` on the `joy_msg` rather than inverting the `trigger_pressed_threshold` value is a clearer demonstration of intention

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.cpp
```diff
@@ -259,6 +272,12 @@ void TeleopDriveJoy::send_drive_command(const sensor_msgs::msg::Joy::SharedPtr j
   angular *= controller_params.scale_angular;
   angular = std::clamp(angular, -controller_params.limit_angular, controller_params.limit_angular);
 
+  if (joy_msg->axes[params_.axis_handbrake] < -params_.trigger_pressed_threshold)
```
tytia: See previous comment on using `std::abs` instead

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.cpp
tytia: Actually the check for the sign of button_index is now not needed and neither is the check for `joy_msg->axes`, as all mapped button callbacks are actually buttons (since you are handling the handbrake in `send_drive_command`)

tytia (changes requested):
Overall you've done a great job, especially for your first PR! I've left some comments for you to address.

Glass2721:
I think I've addressed all the comments. Let me know what you think!

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.cpp
tytia, line 1: lgtm :) 

One last thing I forgot to mention though, could you make it output a message on handbrake activation and deactivation, i.e. going from not pressed to pressed and vice versa?

Also gotta remove the "NOT IMPLEMENTED" part for the handbrake from the print controls string at the bottom

tytia (changes requested):

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.cpp
Glass2721: Everything mentioned should be addressed now. The implementation for the handbrake message is a little jank (I thought that alternatives were either more jank, or a re-write of the node), so if you'd like me to change that I definitely can. Let me know if you want me to do anything else.

tytia: Nah I reckon that's how I would've implemented the handbrake message as well, dw 👍

tytia:
Lgtm, feel free to merge!

tytia (approved):

PR #735, opened 2025-10-12
Terminal Launcher

V01DBREAKER:
## Description
This PR will address a key issue in our current base station setup, it takes too long to open all the terminals needed for our software and autonomous stacks!
This PR contains a structure using nix and some bash magic to quickly fire up a bunch of terminals as well as any commands that may need to be run before and after, essentially allowing us to setup base station with less commands!
Terminals will function exactly as if a user ran them, with the history of the last command and aliases intact in both local and SSH.
The nix structure should be easy to read and extensible so that new setups are easy to create!

This package builds with ws-build, creating launch scripts that are version controlled.
The package also creates a git metadata file so a ws-build has an exact commit + patch you can track it to (if non committed changes are present)
We now just need to create setups (future PR?) but avoid using hard coded paths as the beauty of this setup is that the launch scripts are linked to the binaries as they are created in the bin folder.

## Steps to Test
Read the README.md at `nixfiles/home/macros/launch/README.md`

## Memes
All in one pressure cooker 🔥 
71nX+8t+pKL

Tags for Notion Integration:
tag:software;tag:nixfiles;tag:ready_for_review;tag:qol

PR #749, opened 2025-11-15
fix: nova-git-metadata: don't make builds fail without .git

Redecorating:
When hydra tries to build this it fails because .git is missing. This means hydra doesn't build this and the final workspace package.

Check if .git is missing and if so then just fail silently to ensure it doesn't break the build.

see also: https://github.com/MonashNovaRover/nova/pull/735#discussion_r2442434487

## Memes

Tags for Notion Integration:
tag:bug;tag:software;tag:ci/cd;tag:ready_for_review


Redecorating:
hydra successfully built this pr so it fixes the issue.

FelicityMatt (approved):
lgtm

PR #771, opened 2025-12-06
feat: Auto Stack Launch files and Launch Improvements

V01DBREAKER:
## Description
Creates launch scripts for ARCh and URCh auto stack.
~~No breaking changes, just need a quick look over from Auto :)~~
There might be breaking changes to launch scripts but not the rest of the stack, so just need some auto look over for the auto stack launch part and then ill merge if good :)

## Memes
autostack

## Notes
I also improved the starting path so it should always start in the nix store path of the build which allows you to run the specific binaries that are built with each build command

## Steps to test
1. Build the workspace `ws-build`
2. Run the script of choice `result/launch/###`
URC Autonomous Navigation (AN): an-one, an-two, an-combined, an-mast
ARCh Mapping and Autonomous (MA): ma-one, ma-two, ma-combined.
Note: combined runs one and two on one machine while one and two refer to each laptop. Mast is not included in combined


nixfiles/modules/home/macros/launch/arch/mapping-autonomous.nix
```diff
@@ -11,16 +11,23 @@ let
   one = {
     pre = pre-shell {payload-name=task-name + " one"; need-rover=true;};
     terminals = [
-      {name = "Base:Rviz"; platform=base; cmd="ros2 launch auto_bringup rviz.launch.py";}
+      {name = "Base:Rviz"; platform=base; cmd="./ros2 launch auto_bringup rviz.launch.py";}
```
TJ-Lee-M, line 14: getting "bash: ./ros2: No such file or directory" in some of the generated terminals... but I assume youre not using the full path so that it can use different binaries not just master?

TJ-Lee-M: ( I tried to cd into the bin file before running the .../launch/an-combined just in case but still no worky :/ )

Chetan-Edupalli: Instead of ./ros2 use mros2. That should fix it

V01DBREAKER: @Chetan-Edupalli  no the intention is to use ./ to use the bin folder builds. seems like my fix didnt work, i'll look into it. Thanks for finding that @TJ-Lee-M i'll fix it when i can

Chetan-Edupalli: The launch script also asks for IP address
image

V01DBREAKER: Should be fixed now

Chetan-Edupalli:
The binary path still doesn't work

image

tytia:
I've just tested it as well now, it seems to me the issue is that the path is present on the laptop because that's where it was built but it is assuming that the same path exists on the rover, which is not where it was built :(

tytia:
Nvm, tested it again after building the same derivation separately on both the computer and the rover, seems to be working!

V01DBREAKER:
Merging this cause Chetan was too slow

PR #785, opened 2025-12-09
Fix: can_slueth crash less on resize terminal

Redecorating:
## Description

catch errors caused by unexpected window resize for can slueth. This means it should no longer crash when you resize your terminal.

also a small fix for running from the git tree, before when you run the main script from the git tree although we added the git copy to the python path, any copy existing in the system python path would be found first, we just insert ourselves to the start of the path instead of the end to fix this.

The screen is never cleared tho, so sometimes after resize there's stuff left in the background.

## Steps to Test

`can start vcan1`
`~/nova/src/other/can_sleuth/can_sleuth/main.py taipan`
resize your terminal aggressively 

## Memes

best termnal shape
image

### DIFF
--- a/src/other/can_sleuth/can_sleuth/main.py
+++ b/src/other/can_sleuth/can_sleuth/main.py
@@ -17,7 +17,7 @@ import sys
 if __name__ == "__main__":
     # ensure we can run this from the git tree
     import os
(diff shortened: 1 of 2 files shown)


angelaguinaldo (commented):
I resized aggressively. 👍!

PR #786, opened 2025-12-10
Feat/can sleuth gui

Redecorating:
## Description

a very simple gui (one window with a long multiline string) for the can sleuth output

default is to try gui else do tui output, but you can specify which one you want with -o (or even get both).

the motivation for this is that the tui output blocks the terminal output so it is harder to see errors.

## Screenshots

long:
image

## Steps to Test

```
nova-build -A pkgs.python3Packages.nova-can-sleuth -o /tmp/can_sleuth
can start vcan1

/tmp/can_sleuth/bin/can_sleuth taipan
/tmp/can_sleuth/bin/can_sleuth taipan -o gui
/tmp/can_sleuth/bin/can_sleuth taipan -o tui -o gui
```

to close, either quit the gui window, or do control-c in the terminal

## Memes

image
kinda mad i had to put all the graphical stuff in a different process, and that the Qt people think that if QApplication() fails to initialise, the correct action is to kill oneself with abort instead of raising a python exception :(

### DIFF
--- a/src/other/can_sleuth/README.md
+++ b/src/other/can_sleuth/README.md
@@ -6,6 +6,11 @@ that allows their state to be presented is a readable/useful manner.
 You can either run `main.py` directly, or from the nova workspace shell, run
 `can_sleuth`.
 
+To build and run just can_sleuth instead of the whole workspace:
+```
+nova-build -A pkgs.python3Packages.nova-can-sleuth -o /tmp/can_sleuth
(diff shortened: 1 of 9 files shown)


angelaguinaldo (approved):
I built and checked the gui on my laptop. It works! Tried can sending twitch forward messages with both the tui and gui open. It's all good.

PR #787, opened 2025-12-10
feat: implement radio monitor

### LINKED ISSUE #601: Rover heartbeat
**Skills Involved:** ROS2, Python, TypeScript, React

### **Context**

During competitions we often loose radio connections or connections get slow. We'd like somehow to be able to monitor this connection in the GUI. 

### **TODO**

There are multiple ways to do this, but perhaps the simplest would be to:
- Create a ROS2 node that publishes data every x seconds
  - The message it sends would include a timestamp of when it was sent
- Add something in the GUI that subscribes to this topic and displays how long it has been since the last message it recieved.
  - Could make this look like an heart rate monitior? or just coloured according to how fast/slow the connection is.
- Also have a look at ROS2 topic QOL and see which one should be used.

Feel free to reach out to your STRs or #software-help if you have any questions :)

---

Tags for Notion Integration:
tag:recruit;tag:software;tag:rover

binudakal:
## Description

Radio connections can be unpredictable during comp, so this feature aims to make basic connection health visible in the GUI (#601) . This has been tested remotely using an existing simulator node, but is yet to be tested on the actual radio.   

## ROS

The tester (radio_tester_pub) and actual radio monitor (radio_monitor) nodes both publish a `RadioStatus` message to the topic `/electronics/radio_status`, containing a timestamp along with each connection statistic. This happens _roughly_ every 1 second in the actual monitor, and is a parameter in the tester (default 1 second). 

## Steps to Test

Run the the actual node while on the same network as the Bullet:
`~/Builds/BuildName/bin/ros2 run electronics radio_monitor.py `
or the tester node to simulate:
`~/Builds/BuildName/bin/ros2 run electronics test_radio.py ` with optional parameters frequency, health (strong/weak/intermittent/disconnected), signal, signal_variation. Defaults to strong health.

In the GUI, the radio status button lives in the navbar and shows current connection health. Clicking it will show a tabbed view of graphs for each connection statistic (signal strength, bandwidth, ping).

## Screenshots

Screencast_20251210_222328

## Memes

image

Tags for Notion Integration:
tag:software;tag:ready_for_review


src/ros/nova-gui/nova-gui/src/ros/topics/rosTopic.ts
```diff
@@ -50,4 +50,5 @@ export enum RosTopic {
   // Other Topics
   BATTERY_STATE = "/battery_state",
   ACTIVATED_NODES = "/activated_nodes",
+  RADIO_STATUS = "/electronics/radio_status"
```
FelicityMatt: can we not call the topic electronics? I want to kill that package and rename to `chassis` so can we use `chassis` instead?

src/ros/rover/chassis/electronics/electronics/radio_monitor.py
```diff
@@ -81,7 +82,15 @@ def __init__(self, device):
         
         if self.is_ros:
             super().__init__("radio_monitor")
-            self.publisher = self.create_publisher(RadioStatus, "/electronics/radio_status", 10)
+            self.publisher = self.create_publisher(
+                RadioStatus,
+                "/electronics/radio_status",
+                qos.QoSProfile(
```
FelicityMatt: did this work with the gui? we've had issues in the past where the gui rosbridge qos is different to whatever is publishing and they're not compatible so i'd just want to double check there's no issues there.

FelicityMatt: looks good tho!

src/ros/rover/chassis/electronics/electronics/tests/test_radio.py
```diff
@@ -52,11 +56,34 @@ class RadioTest (Node):
...
+    "intermittent": (-85, 3),
+    }
+
     # Initialises the ROS messages and nodes
     def __init__(self):
         super().__init__('radio_tester_pub')
-        print("Initialising ROS Radio Tester")
 
+        # Declare parameters
+        self.declare_parameter(self.FREQUENCY_PARAM, 1.00)
```
FelicityMatt, line 72: thank you so much for using parameters!!

src/ros/nova-gui/nova-gui/src/components/navbar/RadioStatusModal/RadioStatusButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+};
+
+export function RadioStatusButton() {
+
+  const uiActions = useUIActions();
+  const radioHealth = useRadioMonitor();
+
+  return (
+    <div>
+      <Button
```
FelicityMatt, line 27: Overall I like the button but would it be possible to always keep it the same width? i saw the gif in the pr comment and it irked me a bit hahah. not a huge deal but would be nice!

FelicityMatt (commented):
overall looking really good! I've just reviewed on my phone and left a couple of small comments, i'll try give a proper review and test soon!

src/ros/rover/chassis/electronics/electronics/radio_monitor.py
binudakal: > did this work with the gui? we've had issues in the past where the gui rosbridge qos is different to whatever is publishing and they're not compatible so i'd just want to double check there's no issues there.

It did work when we tested it but this was probably overkill anyway considering its publishing over the same network, I've reverted it to the original default qos just to avoid any issues

FelicityMatt: nice, if it worked that's fine tho

FelicityMatt (approved):
Looks good to me!! I really like the modal design! and thank you for making the button the same width 😂

Issue #699, opened 2025-09-18
Battery Status Telemetry


FelicityMatt:
**Skills Involved:** ROS2, React, Typescript

### **Context**

Part of the ARCh competition requires us to monitor how much battery we are using throughout the task (see battery efficiency multiplier in rules). We'd like to add some telemetry around how full the battery is.

We currently have a GUI component that displays the current battery percentage in the top right.
We also have this node that needs to be tested that listens to the CAN bus and publishes the data it recieves.
https://github.com/MonashNovaRover/nova/pull/194 - It just needs to be tested (Depending on timings, it can probably also be migrated to Python Control2 when that is ready to be used)

In the GUI we'd like to extend the current functionality with functions like "it's consumed this much battery in x amount of time" and pretty graphs displaying how its performed over time would also be cool. Basically knowing statistics around how its performing will be useful.

Something to keep in mind is that pages will be refreshed and this should be persisted, but will need to be reset when opened for the first time at say a new field test.

### **TODO**

- Add pretty graphs / statistics as a drop down/modal for the battery GUI
- Talk to electrical (can msg software STRs for a contact) around what they'd like displayed / what info they will be sending.
- Test ROS2 node (and/or rewrite)

---

Tags for Notion Integration:
tag:recruit;tag:software;tag:nova-gui;tag:rover

jacksonmacc:
Superseded by https://github.com/MonashNovaRover/nova/issues/1237

PR #790, opened 2025-12-11
Feat/battery state publisher

cappatan:
## Description
battery state publisher adds battery_state.py which contains the code for the ROS2 node battery_state, and battery_state.yaml as a parameters file. The parameters file just sets the canbus as can0 instead of the default can1.

## ROS
battery_state is a ROS2 node that publishes messages called BatteryState onto the topic /battery_state, parsing mV and centiamps into volts and amps. It publishes with an update rate of 5 hz.

## Relevant Issues
This PR is part of the solution for the Battery Status Telemetry
[#699](https://github.com/MonashNovaRover/nova/issues/699) issue and can be used to track power usage for the ARCh power usage task.
https://github.com/MonashNovaRover/nova/tree/battery-status-telemetry is where I am currently working on the GUI component that will listen to this topic & allow for actual power tracking.

## Screenshots
Boring screenshots ): 
image
image

## Steps to Test
1.  Run gui-rosbridge
2. In a different terminal, run can start vcan0
3. In a different terminal, run ~/Builds/<build name>/bin/ros2 run electronics battery_state.py --ros-args --params-file ~/nova/src/ros/rover/nova_bringup/params/battery_state.yaml
4. To test values, you can send current and voltage values in hex on canID 4B2. I used this [website](hhttps://www.rapidtables.com/convert/number/hex-to-decimal.html) to convert values.
Messages are formatted as 4B2#AAAABBBB where AAA is the current in centiamps and BBBB is the voltage in mV in hex. The battery node listener converts these values to amps and voltage so AAAA/100 and BBBB/1000 are what gets displayed by python control 2.

e.g. cansend can0 4B2#12345678 sends 4660centiamps and 22136mV, but gets displayed as 46.6 and 22.136 when converted.

## Memes
I cannot think of any memes, but I hope this cute picture of my cat wearing an outfit I made for her makes up for it. ^_^
yuuna

Sorry it took so long to make a PR out of this (its been done for a while .. ). I was nervy >:


src/ros/rover/chassis/electronics/electronics/battery_state.py
```diff
@@ -0,0 +1,117 @@
...
+class BatteryStateHardware(HardwareInterface):
+
+    def __init__(self, contexts: Contexts):
+        super().__init__(contexts)
+        self.logger.info(f"Hardware interface initialised")
+        self.bus = contexts[jcan.Bus]
+
+        self.voltage = float('nan')
+        self.current = float('nan')
+        self.declare_parameter("VOLTAGE_CURRENT_CANID", 0x4B2)
```
FelicityMatt: Can we make this a class variable, we don't like magic strings throughout, this just makes sure that if you need to change this it will propagate and you don't have to change the string in x number of places. Also convention is that parameters are lowercase snakecase :)

eg add it on line 65 like
```py
CANID_PARAM = "voltage_current_canid"
```

Then referencing like:
```py
self.declare_parameter(CANID_PARAM, 0x4B2)

#...
can_id = self.get_parameter(CANID_PARAM).value
```

src/ros/rover/chassis/electronics/electronics/battery_state.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        self.logger.info(f"Received data: {frame.data} from: {frame.id:X}")
+        current = int.from_bytes(frame.data[0:2], 'big', signed=True) 
+        voltage = int.from_bytes(frame.data[2:4], 'big', signed=False) 
+
+        self.parsed_current = current/100.0
+        self.parsed_voltage = voltage/1000.0
+        self.voltage_state.value = self.parsed_voltage
+        self.current_state.value = self.parsed_current
+        self.logger.info(f"current (A): {self.parsed_current}, voltage (V): {self.parsed_voltage}")
```
FelicityMatt: Do you know how often the board will send can messages? just conscious that this could really fill up the terminal if can messages are sent too often in real life

src/ros/rover/chassis/electronics/electronics/battery_state.py
```diff
@@ -0,0 +1,117 @@
...
+        pass
+
+
+if __name__ == '__main__':
+    print("Setting up!")
+    rclpy.init()
+
+    node = Node("battery_state")
+
+    PythonControl(node, update_rate=5, can_bus="can0") \
```
FelicityMatt, line 112: I also imagine that the battery would update super slowly so we may not have to publish 5 times a second but also probably doesn't matter too much.

FelicityMatt (commented):
Looking good! Going to test now

src/ros/rover/chassis/electronics/electronics/battery_state.py
```diff
@@ -0,0 +1,117 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+to run with parameter file:
+$ ~/Builds/<build name>/bin/ros2 run electronics battery_state.py --ros-args --params-file ~/nova/src/ros/rover/nova_bringup/params/battery_state.yaml
+"""
+
+import rclpy
+from rclpy.node import Node, ParameterDescriptor
+from sensor_msgs.msg import BatteryState
+from python_control2 import PythonControl, Controller, Contexts, InterfaceCollection, Interface, HardwareInterface
+from python_control2.hardware_interfaces import CMDHardware
```
Redecorating: I think this import is unused.

Redecorating (commented):
nice

src/ros/rover/chassis/electronics/electronics/battery_state.py
cappatan: added these changes and made info send from self.logger.debug instead of .info, should be ready for 2nd review now ^^

FelicityMatt (approved):
lgtm, merge away

PR #800, opened 2025-12-15
Feat: can_sleuth battery

Redecorating:
## Description

add battery to the can sleuth/sim

also refactor it a bit to make it easier to decode binary bytes from can messages into human readable data. We do this first by using struct.unpack, and then the user supplies a function that converts it to human text (simplest is just `str`)

## Screenshots

image

## Steps to Test

`can_sleuth -e taipan -e drive` there should be battery as well!

## Memes

me when

### DIFF
--- a/src/other/can_sleuth/README.md
+++ b/src/other/can_sleuth/README.md
@@ -47,7 +47,7 @@ Devices
 - LED Driver (not implemented)
 - Kiln (not implemented)
 - Misc Science stuff (not implemented)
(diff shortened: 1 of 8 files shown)


Redecorating:
I was using this to test the battery ros node and that worked fine

Redecorating:
im not sure if we should redo the attributes thing to simplify it (eg make attribute.value not a function and so forth)

PR #802, opened 2025-12-17
[WIP] End effector control

jacksonmacc:
## Description

WIP so Bailey can leave comments, please don't merge!

Adds a new control mode for the end effector. Hold down RB on an xbox controller, use left stick to move the actuator and the face buttons to move the grabby bits (this might change).

Things I need to do still:
* Bailey suggested using a mirror joint for the grabby bit to make both sides move together
* There is a lot of duplicated code between the arm controller and the end effector controller which I need to cut down on
* Add the proper can ids for the cmds
* There is some mysterious jittering when the motors reach their limits

## Changelogs

- Adds ```nova_end_effector_controller``` package
- Modifies ```teleop_arm``` and ```arm_bringup``` to support ```nova_end_effector_controller```
- Modifies ```teleop_arm``` to use the new control scheme
- Modifies arm URDF to include new joints
- Fixes a bug with a gazebo script where "true" wasn't capitalised

## Screenshots

[Screencast From 2025-12-16 13-48-03.webm](https://github.com/user-attachments/assets/549a681f-e158-4fa1-8b0b-28294dcce840)

## ROS

* Adds a new controller, NovaEndEffectorController
* Sends end effector joint commands over the "/arm/end_effector_joint_velocity_command" topic

## Steps to Test

Run:
```bash
```
Open another terminal and run:
```bash
ros2 launch teleop_arm teleop.launch.py teleop_params:=/home/nova/nova/src/ros/rover/arm/teleop_arm/params/teleop.yaml log_inputs:=True
```

## Memes

image (1)


TJ-Lee-M (approved):
tested, lets merge!!
can revisit to adjust speeds from operator feedback after

PR #810, opened 2026-01-05
Feat/camera preset storage

### LINKED ISSUE #212: Save Camera Presets on GUI
**Skills Involved:** Typescript, React

### **Context**

We compete in eight seperate tasks across two compitions, meaning that we have eight different sets of operators. Each task may have the same cameras rotated/located in different ways that mean operators have to always change the filters at the start of every GUI reload.

At the moment we have one static configuration for all cameras. We would like to improve this and have multiple configurations for cameras that operators can choose between.

Secondly, currently when we take a screenshot of a camera feed it is hardcoded what it is named:
https://github.com/MonashNovaRover/nova/blob/15a3254a51bdc6ccbf4c26d7e2ff09482788b5c4/src/ros/nova-gui/nova-gui/src/components/cameras/CameraComponent/CameraComponent.tsx#L105

During competition, especially the science task, we have to take photos of different sites or objects so we want to impart more information onto the image so that it takes less time to sort through the photos.

### **TODO**

- create profiles to load certain preset settings for each camera.
- be able to change settings and then hit "save" and enter a new config name.
- use `useGenericStore` to save the profiles in local storage. **
- be able to pass in a custom function to the camera component that takes in the camera name and time and returns a screenshot name that it should be downloaded as.
- [less pressing] have the ability to select different screenshot presets, (for example we want different names for Delivery vs Science ARC vs Science URC). This is less pressing and I'm not sure how feasible it is so experiement!

** Please consult @FelicityMatt on how this should be stored, it should be done after improvements to `useGenericStore` to remove from a monostorage.

Feel free to reach out to your STRs or #software-help if you have any questions :)

---

  
  Tags for Notion Integration:
  tag:recruit;tag:software;tag:nova-gui

yens5757:
## Description

This PR is to have a camera preset storage, so you can save and load presets/profiles(whatever you wanna call it) that you've set. Ideally, this should work; however, I only tested it with webcams and would need some detailed checking on an actual rover camera. 

Didn't really do a lot of testing on the screenshot part, though it should be able to work as intended.

## Changelogs

Few files I've changed.
1. CameraProfilesState.ts - Define TypeScript types for the camera profile system. Notice that I didn't add a lot of control over what the value should be. I would assume that is the right way, since the UI wouldn't allow the user to have a value that isn't in the range anyway
2. CameraComponent.tsx - Each camera can listen for save/load events, and add custom naming for screenshots, although I didn't test it at all.
3. RootReducer.ts - Generic store for the camera preset
4. cameraProfileEvents.ts - An event system that allows SaveAllCamerasModal and CameraPresetDropdown to talk to all CameraComponent instances without direct coupling.
5. SaveAllCamerasModal.tsx - Save button for saving preset. I've added a 500ms delay to collect the result; however, I am unsure how this would perform on other devices. Maybe we should test whether we want to increase or decrease the delay. However, right now, 500ms works just fine.
6. CameraPresetDropdown.tsx - Add the dropdown for after the button click
7. CamerasPage.tsx - Add the button in the page
## Screenshots

{BE3A92C2-3176-4530-8779-BF59D77B5CC9}

## Steps to Test

Click the Preset button in all camera pages, test if we can save the preset, load the preset, and see if it actually applies to each camera, and see if it shows up only in camera pages. Test the screenshot part as well, see if we can do the different naming.

## Memes

Tags for Notion Integration:


src/ros/nova-gui/nova-gui/src/components/cameras/CameraComponent/CameraComponent.tsx
```diff
@@ -102,14 +160,20 @@ export const CameraComponent = (props: CameraComponentProps) => {
...
+          
+          link.download = filename;
           link.click();
         }
       }
     } else {
       toast("Unable to Take a Screenshot");
     }
-  }, [videoRef, cameraSerial, currentSite]);
+  }, [videoRef, cameraSerial, currentSite, cameraName, props.getScreenshotName]);
```
FelicityMatt: Can you please fix the linting warning

src/ros/nova-gui/nova-gui/src/components/cameras/CameraPresetDropdown.tsx
```diff
@@ -0,0 +1,104 @@
+import {
```
FelicityMatt, line 1: Can we move this file out of the nav bar and into the camera folder (not in the camera component folder though)

src/ros/nova-gui/nova-gui/src/components/navbar/TopBar/CameraPresetDropdown.tsx
```diff
@@ -0,0 +1,104 @@
...
+  return (
+    <Dropdown placement="bottom-end">
+      <DropdownTrigger>
+        <Button
+          size="md"
+          color="primary"
+          variant="ghost"
+          className="w-36"
+        >
+          Preset
```
FelicityMatt: I'd like this button to be an icon not a name, I think it looks better and has more meaning? atm it also has the same if nor more presence than the control panel, which is much more important. I can also see people getting confused between the two when we'd like to be able to visually distinguish between the two buttons.

I had a play around and I like this design, I used the `user` feather icon and removed the fixed width.

Before
image

After
image

src/ros/nova-gui/nova-gui/src/redux/models/CameraProfilesState.ts
```diff
@@ -0,0 +1,34 @@
...
+  timestamp: number; // Unix timestamp when profile was created
+  cameras: {
+    [cameraSerial: string]: CameraSettings;
+  };
+}
+
+/**
+ * Redux state for managing saved camera profiles
+ */
+export interface CameraProfilesState {
```
FelicityMatt, line 26: Can you please add a "last loaded" attribute that specifies the last camera preset that was loaded and always load by default the most recently selected one.

During field tests / comp we often have to reload the page and having to re-select the presets every time will get very annoying. The GUI should remember the last one.

FelicityMatt: It would also be nice to have some visual indicator in the drop down of which preset was last selected, so that you know which one you are on. (this could be a string somewhere of what is selected, or a different background colour of the preset option)

In the case that the user changes a preset, happy to leave it as the last selected preset or ideally to make have no visual indication since none are currently active, but if it's too much hassle don't worry about it.

FelicityMatt (changes requested):
Overall functionality is looking really good!! Just a couple of smaller comments

FelicityMatt:
The gui fails to build

image

FelicityMatt (approved):
lgtm

Please rebase, check it builds and then I'm happy for you to merge it!

Issue #808, opened 2025-12-29
camera usb bandwidth restrictions


Redecorating:
### **Context**

rn we can't find any usb3 cameras so all our cameras on each bus have to share 480mbit/s

documenting what combinations work/don't work here.

### **TODO**

-

Redecorating:
all this testing is with gst-launch autovideosink (so no webrtc).
This means that my latency is less as the encode-webrtc steps
are skipped. Unless otherwise noted, cameras not working is because
of usb bandwidth errors.
```
gst-launch-1.0 v4l2src device=/dev/video6 ! image/jpeg,width=3264,height=2448 ! decodebin ! videoconvert ! autovideosink
```
```
# fish syntax not bash
for node in 2 6 10 14
    gst-launch-1.0 v4l2src device=/dev/video$node ! video/x-h264,width=1920,height=1080,framerate=25/1 ! decodebin ! videoconvert ! autovideosink &
end; read x; kill (jobs -p)
```
etc

4 LifeCams:

4/4: video/x-raw,width=640,height=480,framerate=10/1
2/4: image/jpeg,width=640,height=480,framerate=10/1
2/4: video/x-raw,width=640,height=480,framerate=15/1
1/4: video/x-raw,width=640,height=480,framerate=30/1


4 IADIY usb "3.0" cams:


1/4: video/x-raw,width=640,height=480,framerate=30/1
1/4: video/x-raw,width=800,height=600,framerate=20/1
1/4: video/x-raw,width=3264,height=2448,framerate=1/1
1/4: image/jpeg,width=3264,height=2448,framerate=20/1
1/4: image/jpeg,width=640,height=480,framerate=30/1

Cannot run more than one without usb complaining about not enough bandwidth.

1 iadiy, 1 lifecam:

Lifecam video/x-raw,width=160,height=120,framerate=15/2
IADIY: mage/jpeg,width=640,height=480 fps 30
OK

L: video/x-raw,width=640,height=480,framerate=10/1
I: image/jpeg,width=640,height=480 fps 30
OK

3 Lifecam, 1 IADIY:

2/3 L: video/x-raw,width=160,height=120,framerate=15/2
1/1 I: image/jpeg,width=640,height=480 fps 30


2/3 L: video/x-raw,width=640,height=480,framerate=10/1
1/1 I: image/jpeg,width=640,height=480 fps 30


2/3 L: video/x-raw,width=640,height=480,framerate=10/1
[...]

PR #817, opened 2026-01-08
cameras2: params: use mjpeg and 800x600@30

Redecorating:
can also do
1280x720@15

https://github.com/MonashNovaRover/nova/issues/808#issuecomment-3695998616
this needs a uvcvideo kernel patch that is in the git issue ^

## Steps to Test

run cameras as per normal, make sure you have that patch (orin has this)


FelicityMatt (approved):
Tested on orin today and it worked, LGTM

PR #823, opened 2026-01-12
feat: migrate kiln to PC2

### LINKED ISSUE #743: Kiln PC2 migration and updates
### **Context**

We have existing kiln/heater logic from urc and arc. We have some new but similar requirements from last year and we also want to migrate the code to use python control 2

The biggest change from last year is that we will be using a QCMD to power the kiln, not a custom solution as before.

We also have similar logic for both the kiln for arc and the heaters for URC, so please try to create a python control2 node that could be used for both (you can have a look at the existing urc heater code and urc gui).

### **TODO**

- [ ] Migrate the existing python control kiln code to v2
  - This will involve creation of a NTC temperature sensor hardware inferface, use of the existing QCMD hardware interface, and a new kiln/heater controller that has a set temp service, publishes the current temp to the gui, and tries to maintain the provided temperature.
  - move any interfaces out of `nova-interfaces` and into `science-interfaces`
  - if it doesn't exist talk to @FelicityMatt 
- [ ] Update the GUI to have both kiln and condenser temperature displayed
  -  Please reuse the widget used for the URC heater, if they are seperate widgets please combine (and add config for one/two temperature displays)
  - if you want to make the widget prettier go ahead.

Tags for Notion Integration:
tag:software;tag:science;tag:nova-gui;tag:rover

Glass2721:
## Description

Migrate from existing kiln_server to one that uses PC2. The new node should support the kiln as part of the 2026 science ARC payload, while the previous node remains to support the 2025 payload. Precise configuration and gui will be updated later when more details have been received from science.

resolves #743 

## Changelogs

- add GenericSensorHardware (replaces IntegerSensor, but should be flexible enough for all simple sensors)
- rename kiln_server.py to kiln_old.py
- add kiln.py (with new pc2 controller)
- integrate kiln.py into arc science launch file

## Steps to Test

Setup:
1. `can start vcan1` and run gui with rosbridge (ensure gui is connected)
2. `ros2 launch science_bringup arc.launch.py` (launches kiln.py)
3. open gui -> arc -> space resources -> microscope (has kiln widget)
4. `candump can1`

Should then be able to `cansend can1 012#3FFF` to set kiln temp to 55 degrees (shows on gui), then starting/stopping kiln and adjusting target temp in gui should show changes in whether both heaters are on or not in the candump (ids currently are 0x010 and 0x011, with data being zero = off, non-zero = on)

## Memes

image

Tags for Notion Integration:
tag:software;tag:rover


src/ros/rover/science/science/science/arc/kiln.py
```diff
@@ -0,0 +1,160 @@
...
+
+        self.logger.info(f"HeaterController {self.name} configured")
+
+        return True
+
+    def on_kiln_command(self, request: KilnCommand.Request, response: KilnCommand.Response):
+        self.logger.debug(f"HeaterController {self.name} received KilnCommand request: {request}")
+
+        self.is_on = request.state
+        self.target_temp = request.target
```
Glass2721, line 108: Is it okay to share state with on_update? (in ros2 control, you need to be careful due to race conditions and realtime safety)

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
```diff
@@ -0,0 +1,80 @@
...
+        :param command_interfaces: A collection of Interfaces used to send messages to hardware. Get any command
+        interfaces you need from this, then store them in member variables.
+        :param state_interfaces: A collection of Interfaces containing the current state of the robot. Get any state
+        interfaces you need from this, then store them in member variables.
+        :returns: None or True if configured successfully. False otherwise.
+        """
+        # Update params
+        self.sensor_name: str = self.get_parameter("sensor_name").value
+        self.can_message_id: int = self.get_parameter("can_message_id").value
+        self.sensor_output: str = self.get_parameter("sensor_output").value
```
FelicityMatt: class variables should be defined in the init, you can get the value when declaring the parameter like:

`self.sensor_name: str = self.declare_parameter("sensor_name", self.name).value`

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
```diff
@@ -0,0 +1,80 @@
...
+        :param interpret_data: Translates raw CAN data into sensor outputs (e.g. velocity, temperature)
+        :param sensor_output: What the sensor outputs (e.g. velocity, temperature)
+        """
+        super().__init__(contexts)
+
+        self.bus = contexts[jcan.Bus]
+        self.interpret_data = interpret_data
+        self.last_value: Any = None
+
+        self.declare_parameter("sensor_name", self.name)
```
FelicityMatt: You shouldn't need to declare a name for the sensor, this will be done when you initialise the the python control 2 system and can be accessed at `self.name` I think.

src/ros/rover/science/science/science/arc/kiln_server_pc2.py
```diff
@@ -0,0 +1,150 @@
...
+from python_control2.hardware_interfaces import CMDHardware, GenericSensorHardware
+from science_interfaces.srv import KilnCommand
+from science_interfaces.msg import KilnData
+
+
+class HeaterController(Controller):
+
+    def __init__(self, contexts: Contexts,
+                 thermistors: list[str] = None,
+                 cmd_name: str = "",
```
FelicityMatt: this should not be thermistor or cmd specific, the controller does not care what the hardware is, we should be able to swap it out with another board and hardware and nothing changes. For example we might use different sensors at URC or the future and we don't want to have to rename them all the time.

I would rename to temp_sensors and heaters or heating cartridges or similar.

src/ros/rover/science/science/science/arc/kiln_server_pc2.py
```diff
@@ -0,0 +1,150 @@
...
+        :param data_topic: Topic that KilnData is published to regularly
+        :param publish_rate: Frequency to which KilnData is published to data_topic
+        """
+        super().__init__(contexts)
+
+        if not thermistors:
+            thermistors = [""] # mutable default value
+
+        self.declare_parameter("thermistors", thermistors)
+        self.declare_parameter("cmd_name", cmd_name)
```
FelicityMatt: see above comment

src/ros/rover/science/science/science/arc/kiln_server_pc2.py
```diff
@@ -0,0 +1,150 @@
...
+        self.logger.debug(f"HeaterController {self.name} updated: "
+                          f"{temperatures} temperatures -> {self.heater_cmd.value} heater level")
+
+if __name__ == "__main__":
+    rclpy.init()
+
+    # TODO: replace with actual hardware/firmware configuration on rover
+    node = Node("kiln_server_pc2")
+    PythonControl(node, update_rate=5, can_bus="can1") \
+        .with_controller("kiln_controller", HeaterController,
```
FelicityMatt: You can have a look at the way I've named things [here](https://github.com/MonashNovaRover/nova/blob/07ba0a61b3be5e353488d4c891628c5b08557953/src/ros/rover/science/science/science/auger.py) but I think we don't need to have kiln in front of everything since this whole thing is the kiln system. The hardware names should ideally also not be super specific to hardware, I'd just have this has `controller`, `heater`, `kiln_sensor`, `condenser_sensor` which maybe a comment specifying what type of sensors they are.

This is a preference thing so if you disagree happy to discuss!

src/ros/rover/science/science/science/arc/kiln_server_pc2.py
```diff
@@ -0,0 +1,150 @@
...
+            self.heater_cmd.value = 0.0
+
+        self.logger.debug(f"HeaterController {self.name} updated: "
+                          f"{temperatures} temperatures -> {self.heater_cmd.value} heater level")
+
+if __name__ == "__main__":
+    rclpy.init()
+
+    # TODO: replace with actual hardware/firmware configuration on rover
+    node = Node("kiln_server_pc2")
```
FelicityMatt: don't need to include `pc2`, can also move the old version to the archive or delete it, it won't work with the new setup anyway

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
```diff
@@ -0,0 +1,80 @@
...
+import jcan
+from ..controller_manager.Interface import Interface, InterfaceCollection
+from .HardwareInterface import HardwareInterface
+from ..controller_manager.Contexts import Contexts
+
+class GenericSensorHardware(HardwareInterface):
+
+    def __init__(self, contexts: Contexts,
+                 can_message_id: int = 0,
+                 interpret_data: Callable[[bytes], Any] = lambda x: int.from_bytes(x),
```
FelicityMatt, line 27: does this work with a list of bytes?

src/ros/rover/science/science/science/arc/kiln.py
```diff
@@ -0,0 +1,150 @@
...
+    PythonControl(node, update_rate=5, can_bus="can1") \
+        .with_controller("kiln_controller", HeaterController,
+                         thermistors = ["kiln_thermistor", "condenser_thermistor"],
+                         cmd_name = "kiln_heater",
+                         calculate_reference_temp = lambda l: l[0], # use kiln_thermistor temperature as the current/reference temp
+                         command_service = "/science/kiln_command",
+                         data_topic = "/science/kiln_data") \
+        .with_hardware("kiln_heater", CMDHardware, can_id = 0x051) \
+        .with_hardware("kiln_thermistor", GenericSensorHardware,
+                       can_message_id = 0x4B1,
```
FelicityMatt: also all these should be defined in a parameter file, but I'm working on making it now so will need to wait a day or two

FelicityMatt (changes requested):
some comments but looking good :)

src/ros/rover/science/science/science/arc/kiln.py
FelicityMatt: I believe this is fine, the on_update is called from a ros2 timer, so I assume ros2 deals with that, but this is the only place you write to the variable so it should be fine. Do some testing and if it doesn't break should be fine.

Also I believe the convention is to have the suffix `callback` in a service callback I think, so I would do `kiln_command_callback`.

src/ros/rover/science/science/science/arc/kiln_server_pc2.py
```diff
@@ -0,0 +1,150 @@
...
+        self.kiln_data_publisher = self.node.create_publisher(KilnData, self.data_topic, 5)
+        self.publisher_timer = self.node.create_timer(1 / self.publish_rate, self.publish_kiln_data)
+
+        self.kiln_command_service = self.node.create_service(KilnCommand, self.command_service, self.on_kiln_command)
+
+        self.logger.info(f"HeaterController {self.name} configured")
+
+        return True
+
+    def on_kiln_command(self, request: KilnCommand.Request, response: KilnCommand.Response):
```
FelicityMatt: also make sure to add doc strings and some general block comments :)

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
Glass2721: Yeah, it takes in an argument of type bytes (from jcan.Frame.data) whose value is then obtained (returned)

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
Glass2721: yeah, I do it like this to parameterise the name of the sensor (as by default, it's just `self.name`). Should I get rid of it?

ETA: actually nvm it probably shouldn't be parameterised

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
```diff
@@ -0,0 +1,77 @@
+from collections.abc import Callable
```
FelicityMatt, line 15: Can you please add this doc string to the hardware interface, I'd like all hardware interfaces (and all controllers) to have this going forward.

https://github.com/MonashNovaRover/nova/blob/bfd7c4d934b02470ecf73c41ffad0798855e5fcf/src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/QCMDHardware.py#L1-L15

For hardware interfaces I'd like it to detail what the hardware interface is for/is used for and maybe a couple of small examples if applicable. I'd also like to include the the command/state interfaces so that we can easily compare hardware interfaces and controllers to see if they are compatible, without having to scour the code.

I'm not sure what the technical equivalent term for joint is here, maybe `<sensor>`?

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
```diff
@@ -0,0 +1,77 @@
...
+        :param interpret_data: Translates raw CAN data into sensor outputs (e.g. velocity, temperature)
+        :param sensor_output: What the sensor outputs (e.g. velocity, temperature)
+        :param initial_value: Value output before receiving first CAN messages
+        """
+        super().__init__(contexts)
+
+        self.bus = contexts[jcan.Bus]
+        self.interpret_data = interpret_data
+
+        self.can_message_id: int = self.declare_parameter("can_message_id", can_message_id).value
```
FelicityMatt: I think we call this just `can_id` everywhere else so if we can remain consistent with that that would be good.

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
```diff
@@ -0,0 +1,77 @@
...
+                 interpret_data: Callable[[bytes], Any] = lambda x: int.from_bytes(x),
+                 sensor_output: str = "value",
+                 initial_value: Any = 0):
+        """ Constructor for GenericSensorHardware
+        Creates state interface "name/sensor_output"
+
+        :param contexts: A collection of dependency injection class instances you can index by class type.
+        :param can_message_id: CAN ID of messages from the sensor
+        :param interpret_data: Translates raw CAN data into sensor outputs (e.g. velocity, temperature)
+        :param sensor_output: What the sensor outputs (e.g. velocity, temperature)
```
FelicityMatt: would this be better named `sensor_unit`? or just `unit`? I was a bit confused what this meant at first, and should add this to the header comment.

src/ros/rover/science/science/science/arc/kiln.py
```diff
@@ -0,0 +1,157 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Controls heaters in the kiln using temperature sensors
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+NODE: kiln
+TOPICS:
+    - publisher: /science/kiln_data [KilnData]
+SERVICES:
+	- service: /science/kiln_command [KilnCommand]
+ACTIONS: None
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
FelicityMatt, line 12: Can you add a section talking about what state interfaces it expects

https://github.com/MonashNovaRover/nova/blob/bfd7c4d934b02470ecf73c41ffad0798855e5fcf/src/ros/rover/science/science/science/auger.py#L8C1-L12C46

src/ros/rover/science/science/science/arc/kiln.py
```diff
@@ -0,0 +1,157 @@
...
+                 heaters: list[str] = None,
+                 calculate_reference_temp: Callable[[list[float]], float] = lambda l: max(l),
+                 command_service: str = "",
+                 data_topic: str = "",
+                 publish_rate: int = 5):
+        """ Constructor for HeaterController
+
+        :param contexts: A collection of dependency injection class instances you can index by class type.
+        :param temp_sensors: List of temperature sensor names (as named in their hardware interfaces)
+        :param heaters: List of heater names (uses "heater_name/effort" command interfaces)
```
FelicityMatt:
```suggestion
        :param heaters: List of heater names (uses "<heater_name>/effort" command interfaces)
```

can you add `<>`, just helps to indicate that this is a joint/interface name and not actually the interface `heater_name/effort`

src/ros/rover/science/science/science/arc/kiln.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        super().__init__(contexts)
+
+        # mutable default values
+        if not temp_sensors:
+            temp_sensors = [""]
+        if not heaters:
+            heaters = [""]
+
+        self.temp_sensors: list[str] = self.declare_parameter("temp_sensors", temp_sensors).value
```
FelicityMatt, line 61: can you add some info logging at the bottom of this file, maybe like configured with these two sensor names and these two heater names, reference temp is xxx - just so when we're running we can be confident that it's configured correctly

src/ros/rover/science/science/science/arc/kiln.py
```diff
@@ -0,0 +1,157 @@
...
+
+        self.heater_cmds = [command_interfaces[h + "/effort"] for h in self.heaters]
+        self.temp_sensor_states = [state_interfaces[t + "/temperature"] for t in self.temp_sensors]
+
+        self.kiln_data_publisher = self.node.create_publisher(KilnData, self.data_topic, 5)
+        self.publisher_timer = self.node.create_timer(1 / self.publish_rate, self.publish_kiln_data)
+
+        self.kiln_command_service = self.node.create_service(KilnCommand, self.command_service, self.kiln_command_callback)
+
+        self.logger.info(f"HeaterController {self.name} configured")
```
FelicityMatt, line 95: can you log which interfaces you got

src/ros/rover/science/science/science/arc/kiln.py
```diff
@@ -0,0 +1,157 @@
...
+
+        return True
+
+    def kiln_command_callback(self, request: KilnCommand.Request, response: KilnCommand.Response):
+        """ Update heater control settings when received new KilnCommand
+
+        :param request: new settings for heater control (temperature, on or off)
+        :param response: success or fail at applying new settings
+        """
+        self.logger.debug(f"HeaterController {self.name} received KilnCommand request: {request}")
```
FelicityMatt: make this info

src/ros/rover/science/science/science/arc/kiln.py
```diff
@@ -0,0 +1,157 @@
...
+        .with_controller("controller", HeaterController,
+                         temp_sensors = ["kiln_sensor", "condenser_sensor"],
+                         heaters = ["left_heater", "right_heater"],
+                         calculate_reference_temp = lambda l: l[0], # use kiln_sensor temperature as the current/reference temp
+                         command_service = "/science/kiln_command",
+                         data_topic = "/science/kiln_data") \
+        .with_hardware("left_heater", QCMDHardware, can_id = 0x00) \
+        .with_hardware("right_heater", QCMDHardware, can_id = 0x01) \
+        .with_hardware("kiln_sensor", GenericSensorHardware,
+                       can_message_id = 0x02,
```
FelicityMatt: you don't have to update but just info, usually science IDs are 3 hex digits long and if a board has a can id of 0A0 then when that board sends sensor information to us, it starts with a 4, so it would be 4A0 (and 4A1, 4A2 etc for other sensors) this is the convention with most boards I think.

FelicityMatt (changes requested):
Looking really good!! Just a couple of comments mostly about docs and naming conventions.

Another thing is that you have very few inline comments, that's fine but also just think about other people on the team that may be in their first year or don't have a ton of experience coding (or with complex python stuff) and just try to make sure you write enough comments for anyone to be able to look at it and understand.

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
Glass2721: done with `<name>/<unit>` as the documented state interface as that's literally what is used, but perhaps something else would be better. I can add an example if you want, but I'm not sure how we want to standardise that so an example how would be good.

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
Glass2721: done

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
Glass2721: renamed to `unit` (I agree it's much better)

src/ros/rover/science/science/science/arc/kiln.py
Glass2721: done

src/ros/rover/science/science/science/arc/kiln.py
Glass2721: done

src/ros/rover/science/science/science/arc/kiln.py
Glass2721: done; I haven't added "reference temp" because I don't think state interfaces will have updated yet. Also `target_temp` is initially zero so I haven't added that in the logging.

src/ros/rover/science/science/science/arc/kiln.py
Glass2721: done

src/ros/rover/science/science/science/arc/kiln.py
Glass2721: done

src/ros/rover/science/science/science/arc/kiln.py
Glass2721: interesting. noted for future reference

Glass2721:
> Another thing is that you have very few inline comments, that's fine but also just think about other people on the team that may be in their first year or don't have a ton of experience coding (or with complex python stuff) and just try to make sure you write enough comments for anyone to be able to look at it and understand.

imo the changes made here are really basic so I don't really have anything to say in inline comments that isn't already obvious in code lol. But definitely something I will try to keep in mind with more complex stuff

Also, I think I've addressed all comments so let me know if you want more changes.

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/GenericSensorHardware.py
FelicityMatt: yeah looks good, we can change it in the future if we come up with a better standard.

FelicityMatt (approved):
LGTM! 

I have tested and it all appears to work, great work!

PR #824, opened 2026-01-13
feat: Science Bringup

FelicityMatt:
## Description

Move science launch files to it's own bringup package.

## Changelogs

- Moved science launch files and paramaters
- Updated teleop science to work with joysticks [untested]

## Screenshots

image

## Memes

image


src/ros/rover/science/science_bringup/launch/urc.launch.py
```diff
@@ -0,0 +1,58 @@
+"""
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<Description>
```
FelicityMatt, line 3: This is a placeholder for a new version of the urc launch file cause the current one is a mess

src/ros/rover/science/science_bringup/launch/arc.launch.py
```diff
@@ -30,76 +27,81 @@ def launch_setup(context, *args, **kwargs):
...
+            executable='analysis_arm.py',
             output='screen',
             emulate_tty=True,
             parameters=[
                 science_params,
             ],
         ),
         Node(
-            name='arc_spinny_part',
+            name='kiln_server',
```
Glass2721, line 94: I've renamed `kiln_server.py` to `kiln.py`, but I can change that back if you want.

FelicityMatt: no I think that's fine to keep, but you can change it in your PR so that this still refers to the current kiln ros2 node we have (this file is just copied and slightly reordered)

TJ-Lee-M (approved):
lgtm!

PR #826, opened 2026-01-14
feat: Log ec candumps and add can-utils

FelicityMatt:
## Description

Since ec causes drive to freak out I've added logic to candump into a file that we can look at whenever there is an issue to try and figure out what is going wrong.

Will need to create a `logs` folder in the orins root directory to work.

## Changelogs

- Log candumps when the ec launch script is run
- Remove drive from the ec script
- Add can-utils to the nova desktop

## Memes

image

### DIFF
--- a/src/ros/rover/nova_bringup/launch/ec_rover.launch.py
+++ b/src/ros/rover/nova_bringup/launch/ec_rover.launch.py
@@ -13,19 +13,29 @@ EDITED BY: Tristan Clark, Taaj Street,
     Victor Bartlinski
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 '''
+from datetime import datetime
 from launch import LaunchDescription
-from launch.actions import IncludeLaunchDescription, DeclareLaunchArgument, OpaqueFunction
+from launch.actions import IncludeLaunchDescription, DeclareLaunchArgument, OpaqueFunction, ExecuteProcess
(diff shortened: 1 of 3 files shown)


src/ros/rover/nova_bringup/launch/ec_rover.launch.py
```diff
@@ -13,19 +13,29 @@
...
 
+    timestamp = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    log_file = f"/home/nova/logs/{timestamp}_ec.txt"
+
     return [
+        ExecuteProcess(
+            cmd=[
+                "bash",
+                "-c",
+                f"candump can0 > {log_file}",
```
Redecorating, line 35: maybe add `-t a` to make it include timestamps.

if this ends up taking a lot of disk space, you could also pipe it to gzip to compress it:

`candump can0 | gzip - > file.txt.gz`
then to view:
zcat file.txt.gz | less
to uncompress:
gunzip file.txt.gz # this will create file.txt

Redecorating (approved):
nice.

might be good to have this in general, not just for ec but we can see how it goes for ec first

src/ros/rover/nova_bringup/launch/ec_rover.launch.py
FelicityMatt: We can add this later if we need I think, we will only need to keep track of them in the case that there is an issue, if there isn't we can just delete the file afterwards.

PR #827, opened 2026-01-16
feat: PC2 Auger

FelicityMatt:
## Description

Migrate the auger code to python control2.

Includes the addition of a new launch file for new code, teleop locking capabilities, and the ability to activate and deactivate pc2 systems.

## Changelogs

- migrate auger code
- add qcmd hardware interface
- create new launch file script and params
- add pc2 activation ability
- configure locking functionality to teleop science.

## Steps to test

run:
```
ros2 run science auger.py
```

```
ros2 launch teleop_science teleop.launch.py joystick:=false
```
can play around with params and add `local:=true` to use local directory.

check can output:
```
candump can1
```

## Memes

image

### DIFF
--- a/src/ros/rover/nova_generic/python_control2/python_control2/controller_manager/ControllerManagerBuilder.py
+++ b/src/ros/rover/nova_generic/python_control2/python_control2/controller_manager/ControllerManagerBuilder.py
@@ -3,6 +3,8 @@ import rclpy
 from rclpy.node import Node
 from typing import Type, TypeVar, List, Any, Optional
 from teleop_python_utils import Inputs
+
+from .Activation import Activation
 from .ControllerManager import ControllerManager
 from ..controllers.Controller import Controller
 from ..controllers.DeferredConstructor import DeferredConstructor
@@ -153,6 +155,43 @@ class ControllerManagerBuilder:
         """
         self._cm.contexts[Inputs] = inputs
         return self
+
+    def with_activation_buttons(self, start_active: bool=False, active_button_name: str="", inactive_button_pool_names: list[str]=[""]):
+        """
+        Allows a system to be activatable by adding an activation object to the cm context
+        that controllers can use to conditionally run code.
+
+        Can only have one Activation in a python control2 system.
+
+        :param start_active: Whether to start active or not, defaults to False.
+        :param active_button_name: Name of button that activates.
+        :param inactive_button_pool_names: Name of buttons that deactivate.
+        """
+        # Declare parameters
+        start_active: bool = self._cm.declare_parameter(
+            "active",
+            start_active,
+            'On start node status').value
+        active_button_name: str = self._cm.declare_parameter(
+            "active_button",
+            active_button_name,
+            'Button name that activates the system').value
+        inactive_button_pool_names: list[str] = self._cm.declare_parameter(
+            "inactive_button_pool",
+            inactive_button_pool_names,
+            'list of button name that deactivates the system').get_parameter_value().string_array_value
+
+        # Get button references
+        if Inputs not in self._cm.contexts:
+            raise AssertionError("`.with_teleop` must be called before `.with_activation_buttons`")
+
+        inputs = self._cm.contexts[Inputs]
+        active_button = inputs.get_button(active_button_name)
+        inactive_button_pool = [inputs.get_button(name) for name in inactive_button_pool_names]
+
+        # Create Activation object and add to cm context.
+        self._cm.contexts[Activation] = Activation(active_button, inactive_button_pool, start_active)
+
 
     def spin(self, default_update_rate: float=20, auto_run_rclpy: bool=True) -> None:
         """ Repeatedly updates until the program ends.
--- a/src/ros/rover/nova_generic/python_control2/python_control2/controller_manager/ControllerManagerBuilder.py
+++ b/src/ros/rover/nova_generic/python_control2/python_control2/controller_manager/ControllerManagerBuilder.py
@@ -190,7 +190,8 @@ class ControllerManagerBuilder:
         inactive_button_pool = [inputs.get_button(name) for name in inactive_button_pool_names]
 
         # Create Activation object and add to cm context.
-        self._cm.contexts[Activation] = Activation(active_button, inactive_button_pool, start_active)
+        self._cm.contexts[Activation] = Activation(active_button, inactive_button_pool, self._cm.node, start_active)
+        return self
 
 
     def spin(self, default_update_rate: float=20, auto_run_rclpy: bool=True) -> None:
--- a/src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/Direction.py
+++ b/src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/Direction.py
@@ -1,0 +1,9 @@
+#!/usr/bin/env python3
+
+from enum import Enum
+
+
+class Direction(Enum):
+    """Enum for the different directions of the motors"""
+    POSITIVE = 1
+    NEGATIVE = -1
--- a/src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/QCMDHardware.py
+++ b/src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/QCMDHardware.py
@@ -68,7 +68,7 @@ class QCMDHardware(HardwareInterface):
         # Update params
         self.can_id: int = self.get_parameter("can_id").value
         self.joint: str = self.get_parameter("joint").value
-        self.reversed: int = 1 if self.get_parameter("reversed").value else -1
+        self.reversed: int = -1 if self.get_parameter("reversed").value else 1
 
         self.max_effort = self.get_parameter("max_effort").value
         self.max_effort_can = self.get_parameter("max_effort_can").value
--- a/src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/QCMDHardware.py
+++ b/src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/QCMDHardware.py
@@ -9,8 +9,8 @@ COMMAND INTERFACES:
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 PACKAGE:        python_control2
 AUTHOR(S):      Felicity Matthews
-CREATION:       13/01/25
-EDITED:         13/01/25
+CREATION:       13/01/26
+EDITED:         13/01/26
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 """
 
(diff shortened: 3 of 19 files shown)


src/ros/rover/science/teleop_science/params/arc/joysticks.config.yaml
```diff
@@ -15,82 +15,93 @@ teleop_science:
...
-      "btn_bottom_l6_state",
-      "btn_bottom_l5_state",
-      "btn_bottom_l4_state",
-      "btn_bottom_r3_state",      # Buttons located at the bottom right
-      "btn_bottom_r2_state",      # Numbered Left to Right, First then Second row
-      "btn_bottom_r1_state",      # Ordered top right to left, bottom left to right
-      "btn_bottom_r4_state",
-      "btn_bottom_r5_state",
-      "btn_bottom_r6_state"
+      "l_btn_thumb_u_state",        # Up / Trigger (behind the thumbstick)
```
FelicityMatt, line 20: prefixed with `l_` and `r_` so that the buttons all had unique names.

src/ros/rover/science/teleop_science/params/arc/game_controller.config.yaml
```diff
@@ -53,13 +53,25 @@ game_controller_source:
...
       buttons:
-        sweeper_sweep:
+        lock:
+          from: "BACK"
+        unlock:
+          from: "START"
+        auger_drill_clockwise:
           from: "A"
-        twitch:
+        auger_drill_antclockwise:
```
Glass2721: misspelling anticlockwise?

src/ros/rover/science/teleop_science/params/arc/teleop.yaml
```diff
@@ -20,14 +20,47 @@ teleop_science:
...
+          value: false
+        log_locked:
+          on: "locked/down"
+          type: log
+          message: "Locked"
+        log_unlocked:
+          on: "locked/up"
+          type: log
+          message: "Unlocked"
+
```
Glass2721: alll commands need to be children of `ros__parameters`, not `ros__parameters.control_modes` (as pretty sure they're only associated with teleop node and not this control mode)

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/QCMDHardware.py
```diff
@@ -0,0 +1,120 @@
...
+        :param command_interfaces: A collection of Interfaces used to send messages to hardware. Get any command
+        interfaces you need from this, then store them in member variables.
+        :param state_interfaces: A collection of Interfaces containing the current state of the robot. Get any state
+        interfaces you need from this, then store them in member variables.
+        :returns: None or True if configured successfully. False otherwise.
+        """
+        # Update params
+        self.can_id: int = self.get_parameter("can_id").value
+        self.joint: str = self.get_parameter("joint").value
+        self.reversed: int = 1 if self.get_parameter("reversed").value else -1
```
Glass2721: is this supposed to be `-1 if self.get_parameter("reversed").value else 1` instead?

src/ros/rover/science/teleop_science/params/arc/joysticks.config.yaml
```diff
@@ -15,82 +15,93 @@ teleop_science:
...
     ]
 
     remap:
       axes:
+        # Auger
         auger_actuation:
-          from: "ax_stick_x"
+          from: "r_ax_stick_y"
+        auger_drill_speed:
+          from: "r_ax_slider"
```
Glass2721, line 92: I'm unfamiliar with the slider, but is it correct to assume it's like a joystick and can range from -1 to 1, right?

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/Direction.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python3
+
+from enum import Enum
+
+
+class Direction(Enum):
+    """Enum for the different directions of the motors"""
+    POSITIVE = 1
+    NEGATIVE = -1
```
Glass2721, line 9: Is it better to have this here or separately where required? idk

src/ros/rover/nova_generic/python_control2/python_control2/controller_manager/ControllerManagerBuilder.py
```diff
@@ -154,6 +156,44 @@ def with_teleop(self, inputs: Inputs) -> "ControllerManagerBuilder":
...
+            start_active,
+            'On start node status').value
+        active_button_name: str = self._cm.declare_parameter(
+            "active_button",
+            active_button_name,
+            'Button name that activates the system').value
+        inactive_button_pool_names: list[str] = self._cm.declare_parameter(
+            "inactive_button_pool",
+            inactive_button_pool_names,
+            'list of button name that deactivates the system').get_parameter_value().string_array_value
```
Glass2721, line 182: why `get_parameter_value().string_array_value` instead of just `value`? (maybe I'm missing smth lol)

src/ros/rover/science/science/science/auger.py
```diff
@@ -1,191 +1,121 @@
...
 if __name__ == "__main__":
-    main()
+    print("Setting up!")
+
+    rclpy.init()
+
+    node = Node("auger")
+    inputs = Inputs(node).with_topics("/science/input")
+
+    PythonControl(node, update_rate=5, can_bus="can1") \
```
Glass2721: should `update_rate` be higher or nah? not sure what update rate `auger_old.py` is using

src/ros/rover/science/science/science/auger_old.py
```diff
@@ -0,0 +1,191 @@
...
+            state_id_off=self.DEPTH_NOT_HIT_DATA,
+            publisher = self.depth_publisher,
+        )
+
+        # Create limits
+        self.auger_bottom_limit = LimitSwitchLimit(
+            logger=logger,
+            bus=self.bus,
+            limit_switch=self.bottom_limit_hall_effect,
+        )
```
Glass2721, line 113: do we still need `depth_publisher`and `depth_hall_effect_sensor` in auger pc2 code?

src/ros/rover/science/science_bringup/launch/arc_old.launch.py
Glass2721, line 1: do we want to modify `launch-science-arc` to refer to this file instead of the new launch file? (as its currently incomplete)

Glass2721 (changes requested):
looks really good. Just had a few issues I've left as comments.

src/ros/rover/science/teleop_science/params/arc/game_controller.config.yaml
FelicityMatt: ah yes I fixed it but forgot to push, will update

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/QCMDHardware.py
FelicityMatt: probably, I'll update

src/ros/rover/science/teleop_science/params/arc/joysticks.config.yaml
FelicityMatt: yes. I made a little function in the controller that converts it to [0,1] lol cause I realised it would also reverse the direction. (I have a feeling no one realised this before??)

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/Direction.py
FelicityMatt: I think it will be commonly used, and it was with old python control so I think it's fine to have as a separate thing rather than redefining everytime.

src/ros/rover/nova_generic/python_control2/python_control2/controller_manager/ControllerManagerBuilder.py
FelicityMatt: it gets funky with the default value, I think it thought it was a byte array instead of a string array so this is just to ensure that the type is correct, but it has the same effect.

src/ros/rover/science/science/science/auger.py
FelicityMatt: previously it was 10 times a second, I will up it. https://github.com/MonashNovaRover/nova/blob/6254544669ef946833e7dcc0f3cd11e38f7bad92/src/ros/rover/nova_generic/python_control/python_control/ControllerNode.py#L37

src/ros/rover/science/science/science/auger_old.py
FelicityMatt: that was for urc, we don't need it for arc, and I think we should create a new python control2 file/system to compose the sensor and will be a problem for urc prep :) (it doesn't really affect the auger controller)

src/ros/rover/science/science_bringup/launch/arc_old.launch.py
FelicityMatt: I think we can leave it pointing to the new one, it's pretty easy to just type the launch command and we won't be doing comp sims for a while where the commands matter, but also I'm hoping we can switch to the new one soon anyway. We can change it closer to comp if we need to.

src/ros/rover/science/teleop_science/params/arc/game_controller.config.yaml
FelicityMatt: updated

src/ros/rover/science/teleop_science/params/arc/teleop.yaml
FelicityMatt: RIP that's why it wasn't working.. THANK YOU for pointing that out lmao

Also I've found an issue with the input publisher control mode that I wrote so will have to fix that, it publishes an empty list rather than all zeros when locked :|.

I'll fix that separately, not an issue with this PR.

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/QCMDHardware.py
FelicityMatt: updated

src/ros/rover/science/science/science/auger.py
FelicityMatt: updated

src/ros/rover/science/teleop_science/params/arc/teleop.yaml
FelicityMatt: lmaooo I've already done it
https://github.com/BaileyChessum/teleop_modular/pull/58

want to approve? :point_right: :point_left:

Glass2721: actually might have a look at that pr (not familiar with teleop modular code though) because right now, if there is any input (e.g. auger actuation) and the control mode is locked without releasing that input, then pc2 won't update the input to 0 (so the auger keeps moving). kinda dangerous.

Glass2721 (approved):
lgtm! :)

FelicityMatt:
hydra is broken for everyone but it builds locally :+1:

PR #831, opened 2026-01-18
[WIP] feat: Unity sim

vic-bart:
## Description

Adding Unity Sim to replace Gazebo Sim as the simulation engine for the 2025/26 Design Cycle.

## Memes

Tags for Notion Integration:
tag:autonomous;tag:software;tag:simulation


src/ros/rover/auto/auto_bringup/default.nix
```diff
@@ -104,7 +106,9 @@ buildRosPackage rec {
       usb-cam
       fast-livo2
       fast-calib
-      demo-nodes-cpp;
+      demo-nodes-cpp
+      ros-tcp-endpoint
+      nova-unity-sim;
```
Redecorating, line 117: If unity sim is a dependency of auto bringup but unity sim is only x86_64 and you want auto bringup built for arm as well i think you'll need to make it only add this depenency if we are on x86

nixfiles/packages/other/nova-unity-sim/src/unity_build/build_Data/MonoBleedingEdge/etc/mono/config
Redecorating, line 1: are these gonna be committed or is it just temporary till you get nix to download it

vic-bart: Temporary until we get Nix to download it.

vic-bart: At this point my plan is to add the files as a release to unity-sim, give Nix an ssh key to download from unity-sim, and have it download the release.

src/ros/rover/auto/auto_bringup/default.nix
vic-bart:
```nix
platforms = [ "x86_64-linux" ];
```
This should do it, testing on hydra.

FelicityMatt:
@vic-bart is this still needed?

PR #839, opened 2026-01-22
[WIP] Thrustmaster Arm Input Mappings

### LINKED ISSUE #728: Thrustmaster Input Mappings
### **Context**

Currently we have written new arm software that has only be configured to be controlled by a controller (e.g. XBox). For comp we use the Thrustmaster joysticks, so we need to add configuration to allow the arm to be controlled by joystick inputs.

For this we will use a Joy node ([documentation](https://docs.ros.org/en/ros2_packages/jazzy/api/joy/)) to get inputs from the joysticks.

You will be working within [`teleop_arm`](https://github.com/MonashNovaRover/nova/tree/master/src/ros/rover/arm/teleop_arm), so i'd advise familiarising yourself with the contents.

### **TODO**

- Create config [like this](https://github.com/MonashNovaRover/nova/blob/16741d2082044b369c179b149f5c41cb77e9dbb2/src/ros/rover/arm/teleop_arm/params/teleop.yaml#L138C1-L237C29) for each joystick
  - Have a look at `teleop_modular` [documentation](https://baileychessum.github.io/teleop_modular/).
- Update the [launch file](https://github.com/MonashNovaRover/nova/blob/master/src/ros/rover/arm/teleop_arm/launch/teleop.launch.py) 
  - Run two joy nodes for each joystick and figure out how to differentiate them.
  - Add a parameter to the file to switch which input you use (joysticks vs xbox controller) (and which joy nodes you run)
  - Can look at [launch file documentation](https://docs.ros.org/en/humble/Tutorials/Intermediate/Launch/Launch-Main.html)

---

Tags for Notion Integration:
tag:arm;tag:recruit;tag:software;tag:rover

jacksonmacc:
## Description

Still needs to be tested, please don't merge!

Adds input mappings for the Thrustmasters to the arm controllers.
Closes #728 

## Changelogs

* Seperated the arm teleop.yaml file into teleop.yaml, controller.config.yaml and joysticks.config.yaml.
* joysticks.config.yaml is a brand new file with mappings to the Thrustmasters

## ROS

* Joy nodes communicate over /arm/joy for the controller, as well as /arm/joy/left and /arm/joy/right for the thrustmasters

## Steps to Test

```ros2 launch teleop_arm teleop.launch.py use_joysticks:=true```
and
```ros2 launch arm_bringup control.launch.py``` in another terminal

## Memes

image (2)

### CODE REVIEW

File: src/ros/rover/arm/teleop_arm/params/joysticks.config.yaml [hunk 1]
```diff
@@ -0,0 +1,128 @@
...
+      thrustmaster_left:
+        type: "teleop_modular_joy/JoyInputSource"
+      thrustmaster_right:
+        type: "teleop_modular_joy/JoyInputSource"
+
+thrustmaster_left:
+  ros__parameters:
+    topic: "/arm/joy/left"
+    button_definitions: [
+      "btn_thumb_u_state",        # Up / Trigger (behind the thumbstick)
```
FelicityMatt: I updated these button definitions to add a `r_` and `l_` prefix cause the buttons all get combined and it meant there would be buttons with the same name.

https://github.com/MonashNovaRover/nova/blob/bfd7c4d934b02470ecf73c41ffad0798855e5fcf/src/ros/rover/science/teleop_science/params/arc/joysticks.config.yaml#L20

Not sure if it makes a difference but I thought better to be safe than sorry.

FelicityMatt (commented):
Looking good!

src/ros/rover/arm/teleop_arm/params/joysticks.config.yaml
jacksonmacc: Thank you for the tip! I went through and changed them!

TJ-Lee-M (approved):
woop woop! :D

PR #843, opened 2026-01-27
feat: PC2 chute and tool rotator

### LINKED ISSUE #745: Chute and Spinny Part control
### **Context**

Science has a new chute design, that requires moving up and down. It will use a positional servo, and they want the ability to go to two predefined positions (up, down) and the ability to twitch (move small increments)

There is also an existing "Spinny Part" code that is the bottom of the analysis arm that uses a positional servo, it needs to be able to rotate between three different positions.

### **TODO**

- [ ] Write a controller that takes joystick inputs to move to defined positions and twitch.
- [ ] Write or use a hardware controller for positional servos.
- [ ] Configure this controller and hardware set for use for each use case
  - If you don't think this is possible please reach out to @FelicityMatt .

Tags for Notion Integration:
tag:software;tag:science;tag:rover

binudakal:
## Description

Use PC2 to migrate spinny part and create chute (#745)
Current chute angle limits are dummy values, so will need to be tested.

## Changelogs
- migrate arc_spinny_part.py to pc2 (tool_rotator.py)
- create chute code
- add controller for moving to defined positions and twitching
- add positional servo hardware interface
- add to arc science launch and param files 

## Steps to Test

I've only added the chute to the game controller mapping because of limited space, but they're both in the joystick mapping.
```
ros2 run science chute.py
```
or
```
ros2 run science tool_rotator.py
```
then
```
ros2 launch teleop_science teleop.launch.py joystick:=false
```
and 
```
candump can1
```

## Memes
meme

### DIFF
--- a/src/ros/rover/nova_generic/python_control2/python_control2/controllers/PresetTwitchController.py
+++ b/src/ros/rover/nova_generic/python_control2/python_control2/controllers/PresetTwitchController.py
@@ -1,0 +1,159 @@
+"""
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Controller for systems which take joystick inputs 
+to twitch and move to preset positions. 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+COMMAND INTERFACES:
+  - command_interface   [value between min_angle
+                         and max_angle]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+PACKAGE:        python_control2
+AUTHOR(S):      Binuda Kalugalage
+CREATION:       24/01/2026
+EDITED:         26/01/2026
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+"""
+from typing import Optional, Dict
+
+from ..controller_manager.Interface import Interface, InterfaceCollection
+from ..controller_manager.Contexts import Contexts
+from ..controller_manager.Activation import Activation
+from ..controllers.Controller import Controller
+from teleop_python_utils import Inputs, Button
+
+
+class PresetTwitchController(Controller):
+    rotation_cmd: Interface
+
+    def __init__(
+        self,
+        contexts: Contexts,
+        min_angle: float = 0.0,
+        max_angle: float = 180.0,
+        positions: Dict[str, float] = None,
+        twitch_max: float = 30.0,
+        command_interface: str = "rotation/position"):
+        """ Constructor for PresetTwitchController
+
+        :param contexts: A collection of dependency injection class instances you can index by class type.
+        :param min_angle: Min allowable angle of the system, in degrees.
+        :param max_angle: Max allowable angle of the system, in degrees.
+        :param positions: The system's preset positions. 
+        :param twitch_max: Maximum position the system will twitch, in degrees.
+        :param command_interface: The command interface to write to.
+        """
+
+        super().__init__(contexts)
+        self.logger.info(f"PresetTwitchController -- I have been __init__ialized")
+
+        self.active = contexts[Activation]
+
+        self.command_interface = command_interface
+
+        self.min_angle: float = self.declare_parameter("min_angle", min_angle).value
+        self.max_angle: float = self.declare_parameter("max_angle", max_angle).value
+        self.twitch_max: float = self.declare_parameter("twitch_max", twitch_max).value
+
+        if positions is None:
+            positions = {}
+
+        # Get inputs
+        inputs = contexts[Inputs]
+
+        self.node_name = self.node.get_name()
+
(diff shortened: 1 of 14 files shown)


src/ros/rover/science/science/science/arc/tool_rotator.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+"""
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Controller for the Tool Rotator of the ARC
```
FelicityMatt: Maybe rename Controller to System since there's no controller in this file, same for chute

src/ros/rover/nova_generic/python_control2/python_control2/controllers/PresetTwitchController.py
```diff
@@ -0,0 +1,159 @@
...
+    rotation_cmd: Interface
+
+    def __init__(
+        self,
+        contexts: Contexts,
+        min_angle: float = 0.0,
+        max_angle: float = 180.0,
+        positions: Dict[str, float] = None,
+        twitch_max: float = 30.0,
+        command_interface: str = "rotation/position"):
```
FelicityMatt: I think maybe instead of asking to pass the whole interface name, just use `<hardware_name>/position`, this will always be position but the name can change, and update the header comment.

The purpose of the header comment is to show what units the controller/hardware use so it doesn't work if you pass in the whole interface name.

src/ros/rover/nova_generic/python_control2/python_control2/controllers/PresetTwitchController.py
```diff
@@ -0,0 +1,159 @@
...
+
+        :param command_interfaces: A collection of Interfaces used to send messages to hardware. Get any command
+        interfaces you need from this, then store them in member variables.
+        :param state_interfaces: A collection of Interfaces containing the current state of the robot. Get any state
+        interfaces you need from this, then store them in member variables.
+        :returns: None or True if configured successfully. False otherwise.
+        """
+        # Save references to interfaces
+        self.logger.info(f"Getting {self.command_interface}")
+        self.rotation_cmd = command_interfaces[self.command_interface]
```
FelicityMatt: should add a parameter that sets it to a provided initial value.

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/PositionalServoHardware.py
```diff
@@ -0,0 +1,109 @@
...
+
+class PositionalServoHardware(HardwareInterface):
+    pos_cmd: Interface
+    frame_id: int
+    function_id: int
+
+    def __init__(self, contexts: Contexts,
+                 frame_id: int=0x0A0,
+                 function_id: int=0x01,
+                 angular_range: float=180.0,
```
FelicityMatt: I'm a bit confused about this name, would just `max_angle` work better here?

src/ros/rover/science/science_bringup/launch/arc.launch.py
```diff
@@ -56,6 +66,16 @@ def launch_setup(context, *args, **kwargs):
 
 
         # Misc - Nodes for misc components
+        Node(
+            name='chute',
```
FelicityMatt, line 77: This should be in the CBeam section (includes auger and kiln and everything attached). Maybe Water extraction is a better name?

src/ros/rover/science/science_bringup/params/arc.yaml
```diff
@@ -5,6 +5,21 @@
...
 
+tool_rotator:
+  ros__parameters:
+    can_bus: "can1"
+    controllers:
+      controller:
+        min_angle: 0.0
+        max_angle: 337.41
+        twitch_max: 30.0
+        sweeper: 0.0
```
FelicityMatt: should this be `sweeper_position` if it isn't it should mention something about position.

FelicityMatt (changes requested):
Amazing work!! I've left some small comments.

Also we should have a chat about science joystick mappings, I'm not a huge fan of having the chute and tool rotator being activatable but we'll have to see if we even have enough buttons to not do that.

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/PositionalServoHardware.py
binudakal: I wanted to keep it distinct from `max_angle` (or similar) used in any controllers for systems with positional servos as hardware, with that representing a _chosen_ limit while the servo's _actual_ limit is represented by `angular_range`. But if this is unnecessary/overthinking it I can change it

FelicityMatt: I understand the distinction, I think my issue was the use of "range" when there's only one value? is you change the type to be a list of two ints I think that works better for me. idk range means two to me, but if you disagree thats fine, would `max_value` be ok instead?

binudakal: Yeah makes sense, I've changed it to `angular_limit` which might be better?

PR #845, opened 2026-01-28
chore: Update teleop_modular revision

FelicityMatt:
## Description

Updating teleop_modular to fix the lock issue where input publisher was not sending 0s when locked.

https://github.com/BaileyChessum/teleop_modular/pull/58


TJ-Lee-M (commented):
:+1:

PR #848, opened 2026-01-28
[wip] Feat/oakenc

Redecorating:


Redecorating:
rn the state of it is that it works but opening the camera feeds doesn't work in the gui - it works fine in the webrtc test page tho. When i had a usb camera and a oak camera together it made (i forgot which one) the camera have really bad latency like >5 seconds.

I haven't made the oak do the encoding of other usb cameras on the oak yet but some of the code is there to support that in the future

it needs some work on moving hard coded stuff into params.

i want to look into VBR and how fast we can restart the depthai pipeline if we want to reconfigure it (and if we can hold onto the Device object when we make a new pipeline)

i also need to check if when the oak is unplugged it handles that right.

to test you do
```
~/builds/oak/bin/ros2 launch cameras2 camera_server_launch.py autostart:=true
~/builds/oak/bin/ros2 launch oakenc oak_camera_directory.launch.py
```

the camera serials the oak uses don't match the ones gui expects so you can't see it in the gui

Redecorating:
probably not gonna work more on this for a bit, once cameras3 is a thing hopefully it will be easier

Redecorating:
i think we are getting rid of oaks

PR #850, opened 2026-01-29
feat: CBeam PC2

FelicityMatt:
## Description

Adding a basic actuator controller that converts axis input into effort output.


binudakal:
Looks good, just some tiny comments

PR #862, opened 2026-02-01
novacarrier-flash: init

### LINKED ISSUE #249: NixOS for the Orin
Tags for Notion Integration:
  tag:autonomous;tag:nixfiles

Redecorating:
flash script for nixos on orin novacarrier with correct devicetree. if you flash with this the devicetree will go into the orin module's firmware so it doesn't actually matter if the booted nixos install has the devicetree patches as the linux kernel loads it from the firmware.

1. plug your computer into the usb c port of the devkit.
2. add a wire bridging the FC-REC pin to any GND pin. These pins are under where the orin module sits, on the side opposite the usb and Ethernet.
3. put the orin module in (you can do this before 1 and 2)
4. power on orin
5. run below commands:
```
$ nova-build -A pkgs.novacarrier-flash -o novacarrier-flash
/nix/store/kskjkk46mnhwsxxz1z55ywwv21r5kxr9-initrd-flash-orin-nano-novacarrier-cross
$ sudo ./novacarrier-flash/bin/initrd-flash-orin-nano-novacarrier-cross
```

you may then install nixos.

closes #249


Redecorating:
i think this will fail to build for arm because i think nvidia doesn't compile their flash tools for arm

hacker1024:
Very nice, I will take a closer look some time this week

Redecorating:
```

in job ‘x86_64-linux.novacarrier-flash’:
error:
       … while evaluating the attribute 'novacarrier-flash'
         at /nix/store/gcjfk53njg58lgnqsi26vz4nj51x46im-source/nixfiles/packages/other/default.nix:18:3:
           17|   reolink = callPackage ./reolink { };
           18|   novacarrier-flash = callPackage ./novacarrier-flash { };
             |   ^
           19| } // import ./nova-launch { inherit pkgs; }

       … while calling a functor (an attribute set with a '__functor' attribute)
         at /nix/store/yk9hdl8nybzb4rsd0fkivnsll35k70ik-source-patched/lib/customisation.nix:306:7:
          305|     if missingArgs == { } then
          306|       makeOverridable f allArgs
             |       ^
          307|     # This needs to be an abort so it can't be caught with `builtins.tryEval`,

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: access to URI 'github:nixos/nixpkgs/77ef7a29d276c6d8303aece3444d61118ef71ac2' is forbidden in restricted mode
```
hydra doesn't like flakes-compat but the build is copy-closured and has a root so it shouldn't matter too much.

i'm sure there is a better way to do this than adding patches to jetpack-nixos to make the novacarrier stuff the default but i don't really understand flakes so this is the simplest way i could get it working.

Redecorating:
removed it from hydra misc jobset. we can tidy it up to not use flakes-compat and not patch jetpack nixos as much later.

PR #865, opened 2026-02-01
Disable Dependabot

FelicityMatt:
## Description

Since we are coming up to comp I am disabling dependabot until after the competition period is over.

Good luck to the next STRs who have to deal with this :)

## Memes

image


TJ-Lee-M (approved):
👍

PR #866, opened 2026-02-01
feat: can launch and logging

Glass2721:
## Description

Adds `can.launch.py` to nova bringup, which launches and starts logging on a specified can bus

## Changelogs
- add `can.launch.py`
- modify drive and payload launch files to include `can.launch.py` _(only modifying current launch files)_

## Steps to Test

1. if not on rover: start vcan: `can start vcan0` (`can.launch.py` won't start vcan)
2. run `ros2 launch nova_bringup can.launch.py bus:=can0 bitrate:=250000 log_name:=test`
3. check can sends to can0 appear in `.nova\log\test_can0_YYYY-MM-DD_HH-MM-SS.log`
4. optional: run launch files of other payloads and check if can is start (if not using vcan) and logging is working

Tags for Notion Integration:
tag:software;tag:rover

### DIFF
--- a/src/ros/rover/nova_bringup/launch/can.launch.py
+++ b/src/ros/rover/nova_bringup/launch/can.launch.py
@@ -9,7 +9,7 @@ PROCESSES:
   - candump
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 CREATION:   27/01/2026
-EDITED:     01/02/2026
+EDITED:     02/02/2026
 EDITED BY: Jonathan Jia
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 '''
@@ -71,12 +71,8 @@ def launch_setup(context, *args, **kwargs):
             logger.error(f"Failed to create CAN log directory {log_dir_expanded}: {e}")
 
     # generate log file path
-    time_stamp = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
-    if log_name:
-        log_file_name = f"{log_name}_{bus}_{time_stamp}.log"
-    else:
-        log_file_name = f"{bus}_{time_stamp}.log"
-    log_path = PathJoinSubstitution([log_dir_expanded, log_file_name])
+    log_file =  f"{log_name}_{bus}.log" if log_name else f"{bus}.log"
+    log_path = PathJoinSubstitution([log_dir_expanded, log_file])
 
     can_logger = ExecuteProcess(
         cmd=[
@@ -122,7 +118,7 @@ def generate_launch_description():
         ),
         DeclareLaunchArgument(
             name='log_dir',
-            default_value='~/.nova/log',
+            default_value='~/log',
             description='directory where CAN message log files are stored'
         ),
         DeclareLaunchArgument(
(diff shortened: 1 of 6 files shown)


src/ros/rover/nova_bringup/launch/can.launch.py
```diff
@@ -0,0 +1,137 @@
...
+        try:
+            subprocess.run(['mkdir', '-p', log_dir_expanded],
+                           capture_output=True, text=True, check=True)
+        except subprocess.CalledProcessError as e:
+            logger.error(f"Failed to create CAN log directory {log_dir_expanded}: {e}")
+
+    # generate log file path
+    time_stamp = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    if log_name:
+        log_file_name = f"{log_name}_{bus}_{time_stamp}.log"
```
FelicityMatt: After discussing with @TJ-Lee-M we thought it might be fine to only have one log file, and just keep overwritting it since we don't need to keep historical logs, only if something goes wrong or if there is a need. In which case we can manually rename to keep safe, thoughts on this?

src/ros/rover/nova_bringup/launch/can.launch.py
```diff
@@ -0,0 +1,137 @@
...
+            description='bitrate of the CAN bus (usually 250000 or 200000)'
+        ),
+        DeclareLaunchArgument(
+            name='log_can',
+            default_value='true',
+            description='log CAN messages sent on the CAN bus'
+        ),
+        DeclareLaunchArgument(
+            name='log_dir',
+            default_value='~/.nova/log',
```
FelicityMatt: does this work? I had issues with this and had to use
`/home/nova/logs`

FelicityMatt (approved):
LGTM some optional comments

src/ros/rover/nova_bringup/launch/can.launch.py
FelicityMatt: I ran it and it's logging to:
/home/nova/.nova/log/ec_can0_2026-02-02_10-22-46.log

why .nova??

Glass2721: Was there any error/warning output associated with creating the log directory or mkdir? .nova should be a hidden folder (so you can't see it just from ls but can cd into it)

I don't really mind if we store it here or in /home/nova/log up to you

Glass2721: .nova cause ros logs at .ros/log; tbh any default is fine let me know which one you prefer

src/ros/rover/nova_bringup/launch/can.launch.py
Glass2721: Would this or keeping logs from e.g. the last 7 days be preferred?

TJ-Lee-M: I think just the previous one log is better - and I would prefer keep the time stamp out of the log name

Glass2721: cool I'll change it to just keeping the latest log file

src/ros/rover/nova_bringup/launch/can.launch.py
TJ-Lee-M: I would rather it not be hidden, /home/nova/log sounds good to me

PR #868, opened 2026-02-01
Feat/novacarrier

Redecorating:
## Description

to make a nixos installation on a novacarrier work, add the following extra stuff:
```nix
  devices.jetson.orin-nano.enable = true;
  devices.jetson.devkit.enable = true;

  nova.networking.rover = {
    enable = true;
    hostname = "novacarrier2";
    wifiIfName = "wlP1p1s0";
    ethernetIfName = "enP8p1s0";
    ethernetIpAddr = "10.0.0.12";
  };
```

this will make a wifi hotspot "novacarrier2-hotspot" that is bridged to the ethernet.

please please please change "novacarrier2" and "10.0.0.12" to not be the same - i want to put a sticker on each novacarrier pcb with numbers 1 2 3 and then a sticker on the ssd with the ip address

the operating system is an attribute of the ssd, not the orin module!

## Memes
sorry for force pushing your branch brandon


Redecorating:
wifi hotspot doesnt work but it doesn't impact ethernet so we can fix it later

rotaryViper (approved):
Sending it

PR #874, opened 2026-02-03
arm joystick mapping and joint speed changes

TJ-Lee-M:
## Description
Remapped joysticks as per operator feedback
Now left jostick is base (twist=J1, x=J3, y=J2), right joystick is wrist (twist=J4, x=J5, y=J6)
Direction of J1,2,3,5 is correct, but J4,J6 were not running and may need reversing later

Speeds were adjusted to operator preference

## Steps to Test
Tested on Monday 2nd Feb during operator training


FelicityMatt (approved):

src/ros/rover/arm/teleop_arm/params/teleop.yaml
```diff
@@ -50,19 +51,21 @@ teleop_arm:
...
       enter_end_effector_mode:
         on: "end_effector_mode/down"
         type: switch_control_mode
-        to: end_effector_mode
+        activate: ["end_effector_mode"]
       enter_joint_space_from_end_effector_mode:
         on: "end_effector_mode/up"
         type: switch_control_mode
-        to: joint_space_mode
+        activate: ["joint_space_mode"]
```
BaileyChessum: this is dangerous -- reduce to two simple cases for the joint and task spaces

src/ros/rover/rover_description/banksia_arm/urdf/arm.ros2_control.xacro
```diff
@@ -22,7 +22,7 @@
 					<param name="integrate_velocity">false</param>
 					<param name="max_position">3.141592654</param>
 					<param name="position_offset">0.0</param>
-					<param name="max_velocity">1.779309114</param>
+					<param name="max_velocity">2.3</param>
```
BaileyChessum, line 25: can we had comment, maybe for the first set of params

this param means, 
the radians per second of the max can command the blcmds accept.

the max radians per second that the blcmds can accept via a CAN command

src/ros/rover/arm/teleop_arm/params/teleop.yaml
```diff
@@ -97,11 +100,11 @@ end_effector_mode:
       actuator_head:
         name: "actuator_head"
         input_name: "actuator_head"
-        scale: 0.1
+        scale: 0.5
```
BaileyChessum, line 99: can we add a comment to what these scales mean

this is the radians per second value that is sent to the controller at full joystick speed.

1 = 1 radian per second sent to the controller which is sent to the hardware interface etc (excluding limits) which theoretically, assuming the hardware interface is configured correctly, moves the joint at 1 radian per second (in joint space)

src/ros/rover/rover_description/banksia_arm/urdf/arm.ros2_control.xacro
```diff
@@ -22,7 +22,7 @@
 					<param name="integrate_velocity">false</param>
 					<param name="max_position">3.141592654</param>
 					<param name="position_offset">0.0</param>
-					<param name="max_velocity">1.779309114</param>
```
BaileyChessum, line 25: 1.7... is the reasonable value for a blcmd+r80 with a gearbox ratio of 100:1. That's only really the base joints

src/ros/rover/arm/teleop_arm/params/teleop.yaml
TJ-Lee-M: not configured in hardware interface atm - we can update comments when we change/merge that (i know, im sorry - but I think we have bigger priorities atm than working that out and I want to merge this for operator training at least)

src/ros/rover/rover_description/banksia_arm/urdf/arm.ros2_control.xacro
TJ-Lee-M: im confused about this...
If max_velocity is the value corresponding to max can msg, then increasing max_velocity would just decrease what we're sending?
We've been changing it as if it is the velocity limit that we need to increase to go faster...

src/ros/rover/arm/teleop_arm/params/controller.config.yaml
```diff
@@ -81,8 +81,8 @@ joy_input_source:
           from: LEFTY
         left_slide:
           from_buttons:
-            positive: "Y"
-            negative: "A"
+            positive: "LEFTSHOULDER"
+            negative: "RIGHTSHOULDER"
```
BaileyChessum, line 85: no :(

Left shoulder is to enter task space

src/ros/rover/arm/teleop_arm/params/controller.config.yaml
```diff
@@ -115,6 +115,6 @@ joy_input_source:
         unlock:
           from: "START"
         task_space:
-          from: "LEFTSHOULDER"
+          from: "TRIGGERLEFT"
```
BaileyChessum: TRIGGERLEFT is not a button

BaileyChessum (commented):
thrustmaster stuff looks good. I don't trust what you did to the controller mapping. See feat/taipan for my opinion on the mapping.

src/ros/rover/arm/teleop_arm/params/controller.config.yaml
TJ-Lee-M: want to remap so EE mode isnt needed (would rather have intuituve "opposite" buttons for pos/neg)

src/ros/rover/arm/teleop_arm/params/controller.config.yaml
TJ-Lee-M: true whoops
I dont like the other button options for this, can we use triggerlieft as an axis or does this need to be a button?

src/ros/rover/arm/teleop_arm/params/teleop.yaml
TJ-Lee-M: (otherwise would love for you to comment interfaces like this all at once - I just dont want to add comments one by one as we come across them when we know that we cant actually use them yet becasue they arent all configured/identified)

src/ros/rover/rover_description/banksia_arm/urdf/arm.ros2_control.xacro
TJ-Lee-M: how did you get to 1.7...?

src/ros/rover/rover_description/banksia_arm/urdf/arm.ros2_control.xacro
TJ-Lee-M: will leave for now to get merged but we should really change this convention to be clearer/safer

BaileyChessum (approved):
nice 

Image

src/ros/rover/rover_description/banksia_arm/urdf/arm.ros2_control.xacro
TJ-Lee-M: measured by trial & error - but knowing gearbox ratio should allow us to roughly calculate other joints

PR #877, opened 2026-02-04
Diff Wrist URDF

### LINKED ISSUE #740: Updated URDF
### **Context**

The arm mount has been has redesigned by arm! We need to update our rviz sim with the new urdf

### **TODO**

-

Tags for Notion Integration:
tag:arm;tag:simulation

V01DBREAKER:
I actually ended up completely redoing the CAD and then fully reexporting the URDF :)

## Description
Closes #740

View with `ros2 launch arm_bringup urdf.launch.py rviz:=True old_arm:=False arm:=True`

I put in the mimic joints but haven't tested if their multipliers are correct on each mimic joint (it should be either 1 or -1 like pre-diffwrist taipan)
Also the kinematic frames might not be in the correct locations (found in `arm.urdf.xacro`) *IMPORTANT FOR IK*

image
I couldn't get a controller working on wsl so i'll need someone to test if all the ros2 control stuff works. 
Not sure how to run it but i think it should just be control, teleop_arm and rviz. (Ask @BaileyChessum )

Note pre-diffwrist taipan had these mimic joints with corresponding multipliers and arrows pointing at where mimic joints were mimicking:
image

I also renamed banksia_arm to taipan and waratah_arm to old_arm

cad model: https://monashnova.onshape.com/documents/226e4acb3f4645fcb2c5e12b/w/3cccc2d52885999e7ab893cc/e/def70ee139d74fbdebd0e1f4

i also updated notion page but i need to update it again cause i learnt it actually exports subassemblies and not just parts meaning i didnt have to boolean all the subparts.
collisions still had to be done with openscad

Arm URDF be like:
image

Tags for Notion Integration:
tag:arm;tag:ready_for_review


jacksonmacc:
Issues still pending:

- End effector doesn't give positional feedback because of the qcmd hardware interface, but TF expects there to be positional feedback. As a result, the end effector turns white and doesn't move. This is also a problem on master. I can't find a way to just get it to ignore the end effector yet.
- I am not sure if the mimic joints are correct, but the arm looks like it is moving correctly (sans end effector).

jacksonmacc:
To test:

```ros2 launch arm_bringup mock.launch.py rviz:=true```

and in another terminal

```ros2 launch teleop_arm teleop.launch.py joysticks:=false```

PR #879, opened 2026-02-04
nixfiles: don't wrap qt apps twice

Redecorating:
avoid wrapping everything in the nova workspace with qt's env vars. this saves about 15-20 seconds in ws-build. the apps using qt already have the qt env vars set in their individual package's wrapping step.


vic-bart:
Built on local, hydra works too, LGTM :+1:

PR #889, opened 2026-02-07
feat: Update aliases and rover-help

FelicityMatt:
## Description

Update aliases with new arm and science commands, and update rover help to reflect the new alias and commands we are using.

## Memes

image

Tags for Notion Integration:
tag:software;tag:ready_for_review

### DIFF
--- a/nixfiles/doc/rover-help.md
+++ b/nixfiles/doc/rover-help.md
@@ -53,7 +53,7 @@ Try the `launch-arm` alias, otherwise use the command below.
 ~/Builds/master/bin/ros2 launch arm_bringup control.launch.py
 
 ## Arm Teleop
-Launch this on the metabox (operator station).
+Launch this on the metabox (base station).
 Try the `launch-teleop-arm` alias, otherwise use the command below.
 ~/Builds/master/bin/ros2 launch teleop_arm teleop.launch.py
 
@@ -64,7 +64,7 @@ Try the `launch-ec` alias, otherwise use the command below.
 ~/Builds/master/bin/ros2 launch nova_bringup ec_rover.launch.py
 
 ## C&E Teleop
-Launch this on the metabox (operator station).
+Launch this on the metabox (base station).
 Try the `launch-teleop-ec` alias, otherwise use the command below.
 ~/Builds/master/bin/ros2 launch teleop_ec teleop.launch.py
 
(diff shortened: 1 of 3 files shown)


nixfiles/modules/home/macros/default.nix
```diff
@@ -89,12 +90,16 @@ in
           # Launch rover or payloads
           rover-help = "more ${cfg.nixfileDir}/doc/rover-help.md";
           launch-teleop = "~/Builds/master/bin/ros2 launch teleop_drive_joy teleop.launch.py";
+          launch-teleop-drive = "~/Builds/master/bin/ros2 launch teleop_drive_joy teleop.launch.py";
```
FelicityMatt, line 93: for if people get confused about what just `launch-teleop` is launching

nixfiles/doc/rover-help.md
```diff
@@ -49,33 +49,53 @@ In another terminal:
...
+Try the `launch-base` alias, otherwise use the command below.
+~/Builds/master/bin/ros2 launch nova_bringup base.launch.py
+
 
 ---------------------------
 
 # Cameras
 Launch this on the jetson
-Replace '?' with either 'arm', 'ec', or 'arc_science'.
+Replace '?' with either 'arm', 'ec', 'science-arc' or 'science-urc'.
```
TJ-Lee-M, line 98: should we make science-arc sr? to match ec

nixfiles/modules/home/macros/default.nix
TJ-Lee-M: I think get rid of laucnh-teleop then? I dont want two alias' for the same thing

TJ-Lee-M (approved):
couple suggestions but just preference - otherwise lgtm

nixfiles/modules/home/macros/default.nix
FelicityMatt: I might consult auto then, I'm worried other things might depend on it, and my intention was to keep it but kind of deprecate it.

nixfiles/doc/rover-help.md
FelicityMatt: possibly, although might be best to do in a different pr cause that'll need to rename the camera param directory and stuff and I can't be bothered lol

PR #891, opened 2026-02-09
fix can bus names on nixos

Redecorating:
## Description

this is built with nixos-rebuild on the nixos ssd in the rover

tested and it does set the names of the can busses so it should be consistant now.


hacker1024 (approved):
Image

PR #893, opened 2026-02-09
gui: nix: filter out node_modules, dist

Redecorating:
these folders make gui-shell rebuild a different nova-gui-modules derivation even if the actual source is the same. this means that it doesn't get the build from hydra and is slower. filtering out these makes it not do that

### DIFF
--- a/src/ros/nova-gui/nix/packages/gui/default.nix
+++ b/src/ros/nova-gui/nix/packages/gui/default.nix
@@ -34,7 +34,7 @@ mkYarnPackage {
   src = builtins.path rec {
     name = "gui";
     path = ../../../nova-gui;
-    filter = lib.novaSourceFilter [ ] path;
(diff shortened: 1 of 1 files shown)


Redecorating:
also faster than gui yarn install:
gui-shell

ln -sf $node_modules/.* /home/nova/nova/src/ros/nova-gui/nova-gui/node_modules/
ln -sf $node_modules/* /home/nova/nova/src/ros/nova-gui/nova-gui/node_modules/

note that we can't just link the whole node_modules folder as vite expects to be able to write to it :(

FelicityMatt (commented):
lgtm, I don't really know what's going on but I ran the gui and it works fine.

Also we run `yarn install` so rarely that I don't really care how long it takes, improving `gui-shell` tho is useful, thanks!

PR #896, opened 2026-02-10
Sweeper PC2

### LINKED ISSUE #838: Migrate sweeper to pc2
### **Context**

Sweeper needs to be migrated to pc2

### **TODO**

- Write a sweeper controller
- Update teleop configs
- Use a jonocard effort hardware interface (write it if it doesn't exist yet)

Tags for Notion Integration:
tag:software;tag:science;tag:rover

rotaryViper:
## Description

It should sweep?

To do:
1. Figure out if 16 bit conversion in sweeper.py is necessary.


src/ros/rover/science/science/science/arc/sweeper.py
```diff
@@ -0,0 +1,124 @@
...
+        """ Constructor, deferred until the control manager has been spun.
+        If you override this method, and want to add your own arguments, just make sure contexts is the FIRST arg
+
+        :param contexts: A collection of dependency injection class instances you can index by class type.
+        """
+        super().__init__(contexts)
+        self.logger.info(f"SweeperController -- I have been __init__ialized")
+
+        self.sweeper_direction = Direction.POSITIVE
+        self.active = contexts[Activation]
```
FelicityMatt: We don't need the sweeper to be activatable, it should always be active. We only really need this when we run out of buttons or we want to use the main joystick axis for more than one thing.

src/ros/rover/science/science/science/arc/sweeper.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # Get inputs
+        # Sweeper axis
+        self.sweeper_axis_name = self.declare_parameter("sweeper_axis", "sweeper_sweep").value
+
+        inputs = contexts[Inputs]
+        self.sweeper_axis = inputs.get_axis(self.sweeper_axis_name)
+        self.sweeper_axis.value = self.to_int16(self.sweeper_axis.value)
```
FelicityMatt: you should not be assigning the sweeper axis value, this value is updated to align with what teleop publishes. The controller should not have to worry about unit conversion like this. The controller should just be sending effort values (between -1 and 1) to tell the sweeper hardware what effort to move at.

src/ros/rover/science/science/science/arc/sweeper.py
```diff
@@ -0,0 +1,124 @@
...
+                self.sweeper_direction = Direction.NEGATIVE
+            case _:
+                self.logger.error(f"Updated to invalid sweeper direction: {new_direction}")
+
+        self.logger.info(f"Updated sweeper direction: {"CLOCKWISE" if self.sweeper_direction == Direction.POSITIVE else "ANTICLOCKWISE"}")
+
+        # Convert to 16 bit integer
+        self.sweep_cmd.value = self.to_int16(self.get_sweep_speed() * self.sweep_direction.value)
+
+    def get_sweeper_speed(self) -> int:
```
FelicityMatt: we don't need this function, the sweeper axis value will be between -1 and 1, the pos/neg part gives direction, and the value speed (effort)

FelicityMatt: in fact since you'll be getting your input from an axis, you don't need the direction attribute at all, that is more for buttons, or things that can change direction with a button.

src/ros/rover/science/teleop_science/params/arc/joysticks.config.yaml
```diff
@@ -70,6 +73,9 @@ thrustmaster_left:
           from: "l_btn_bottom_r5_state"
         tool_rotator_nir_probe:
           from: "l_btn_bottom_r6_state"
+        # Sweeper
+        activate_sweeper:
```
FelicityMatt: don't need see joystick mappings:

Image

FelicityMatt (changes requested):
looking good so far, a couple of comments.

Also since the thumb axis is only -1 or 1, it would be good to add the slider as a speed axis to allow the sweeper to change speed as needed.

src/ros/rover/science/teleop_science/params/arc/joysticks.config.yaml
```diff
@@ -52,6 +52,11 @@ thrustmaster_left:
         # Tool Rotator
         tool_rotator_speed:
           from: "l_ax_slider"
+        # Sweeper
+        sweeper_actuation:
+          from: "l_ax_thumb_y"
+        sweeper_speed:
+          from: "l_ax_slider"
```
Glass2721, line 59: the slider (unintuitively) is an axis which can take values from -1 (bottom) to 1 (top). so you'll probably have to remap this range from [-1, 1] to [0, 1] (so direction is controlled only by the thumb button slider thing). It isn't documented well but you can do it like this using `range`:
```
        scraper_speed:
          from: "r_ax_slider"
          range:
            in: [-1.0, 1.0]
            out: [0.0, 1.0]
```

Glass2721: @FelicityMatt left or right slider for sweeper speed?

src/ros/rover/science/science/science/arc/sweeper.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self, contexts: Contexts):
+        """ Constructor, deferred until the control manager has been spun.
+        If you override this method, and want to add your own arguments, just make sure contexts is the FIRST arg
+
+        :param contexts: A collection of dependency injection class instances you can index by class type.
+        """
+        super().__init__(contexts)
+        self.logger.info(f"SweeperController -- I have been __init__ialized")
+
+        self.active = contexts[Activation]
```
Glass2721: probably remove `self.active = contexts[Activation]` because you're not using it

src/ros/rover/science/science/science/arc/sweeper.py
```diff
@@ -0,0 +1,91 @@
...
+        # Get inputs
+        inputs = contexts[Inputs]
+
+        # Sweeper actuation
+        self.sweeper_axis_name = self.declare_parameter("sweeper_axis", "sweeper_actuation").value
+        self.sweeper_axis = inputs.get_axis(self.sweeper_axis_name)
+
+        # Sweeper speed
+        self.speed_axis_name = self.declare_parameter("speed_axis", "sweeper_speed").value
+        self.speed_axis = inputs.get_axis(self.speed_axis)
```
Glass2721: I think you mean `inputs.get_axis(self.speed_axis_name)` instead

Glass2721 (changes requested):
Almost there! Tested with fixes using game pad and works. I've left some comments for issues I've found and asked felicity for stuff I'm not sure about.

@FelicityMatt are we using jonocard v2 or qcmd? and do we want to have game pad controls for this or nah?

src/ros/rover/science/teleop_science/params/arc/joysticks.config.yaml
FelicityMatt: same side as buttons, so left I think

FelicityMatt:
This will be on jonocard, should just be 2 hex digit signed magnitude, and yeah probs not worth adding game pad controls, we can add them locally if we ever need to test with it.

src/ros/rover/science/science/science/arc/sweeper.py
rotaryViper: oopsies

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/ContinousServoHardware.py
```diff
@@ -0,0 +1,120 @@
...
+    joint: str
+    reversed: bool
+    max_effort: float
+    max_effort_can: int
+
+    def __init__(self, contexts: Contexts,
+                 joint: str="",
+                 can_id: int=0,
+                 reversed: bool=False,
+                 max_effort: float=1.0, max_effort_can: int=0xFF):
```
Glass2721: Instead of 0xFF, should `max_effort_can = 0x7F` instead as its signed?

src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/ContinousServoHardware.py
```diff
@@ -0,0 +1,120 @@
...
+        :param period: The time elapsed since the last update, in seconds.
+        """
+        pass
+
+    def on_write(self, now: float, period: float):
+        """ Called to write to hardware using values stored in command interfaces.
+        :param now: The current time, in seconds
+        :param period: The time elapsed since the last update, in seconds.
+        """
+        if self.effort_cmd.value != 0:
```
Glass2721: It might be better to send zero effort anyways, wdyt?

Glass2721 (changes requested):
A few more comments :). After this, rebase and fix merge conflicts, and I'll test before approving.

Glass2721 (approved):
tested and works! ty for your work on this

rotaryViper:
tysm :D

PR #897, opened 2026-02-12
fix: drive telemetry (and game pad rumble)

### LINKED ISSUE #807: Update drive GUI
### **Context**

With the shift to new drive our gui drive widgets are no longer compatible, and no longer work. This means that operators are no longer able to see what drive mode they are in or if the controller is connected or locked.

We need to check all of the widgets and ensure that they are working as expected. This includes the drive modes widget, the wheel effort ones and a speed one?

### **TODO**

- Check all drive components in the gui and figure out what needs to be fixed.
- Update the drive modes widget
  - Add a button/option for holonomic drive
  - Create a teleop control mode that just publishes a boolean message saying whether the system is locked or not (can probably be in the teleop repo, can have a look at input publisher there for similar code (this will be a lot simpler tho))
  - update the gui to subscribe to this topic, and remove any old ros stuff.
  - Investigate what different control modes publish when they switch and see if any are useful to update the GUI display, otherwise talk to bailey and see what can be done. It would be ok to do something less ideal now and create a new issue if it requires a larger change.
- Fix anything else that is broken

Glass2721:
## Description

Enables and fixes drive telemetry for the current state of new drive, and adds a feature where sufficiently high wheel effort (current) rumbles the game pad to warn operators and give them better situational awareness.

resolves #807 

## Changelogs

- make gui use joint state broadcaster instead of topic published by old drive
- add ackermann drive to gui
- configure joint state broadcaster to publish effort state interfaces (with addition of said interfaces to the urdf)
- make teleop drive joy publish to drive_info when teleop state is updated (including when game pad is disconnected via a timer)
- add ros__parameters (not generate parameter library) for teleop drive joy so params can be modified without rebuilding by modifying this file and passing local:=true to teleop.launch.py 
- add game pad rumble feature via joint state callback in teleop drive

## Steps to Test

Tested on rover and works; rumble is currently disabled but you can test with rumble_enable = true in teleop.yaml.

1. run drive, teleop drive (via vcan; could not get gazebo to output effort) and gui
2. report max current to front left wheel with cansend can0 411#7FFF7FFF (change 411 to 4?1 where ? is the id of the drive blcmd)
3. see changes in gui (exceeding 100% is expected, as we assume 0.45 is max, whereas effort could be reported up to 1; serves as effective "max")
4. if rumble_enable = true in teleop.yaml: controller should be rumbling

### DIFF
--- a/src/ros/nova-gui/nova-gui/src/components/drive/DriveModeWidget/DriveModeWidget.tsx
+++ b/src/ros/nova-gui/nova-gui/src/components/drive/DriveModeWidget/DriveModeWidget.tsx
@@ -67,7 +67,7 @@ const DriveModeWidget: React.FC<IDriveModeWidgetProps> = (
       </CardHeader>
       <CardBody className="flex justify-center flex-col content-center">
         <div>
-          <div className="grid grid-flow-col gap-3 auto-cols-fr">
+          <div className="grid grid-flow-row grid-cols-[repeat(auto-fit,_minmax(10em,_1fr))] gap-3">
             {driveModes.map((mode, index) => (
               <DriveModeButton
                 key={index}
--- a/src/ros/rover/drive/drive_bringup/launch/drive.launch.py
+++ b/src/ros/rover/drive/drive_bringup/launch/drive.launch.py
@@ -85,7 +85,9 @@ def launch_setup(context, *args, **kwargs):
                 Node(
                     package='controller_manager',
                     executable='spawner',
-                    arguments=['joint_state_broadcaster'],
+                    arguments=['joint_state_broadcaster',
+                               '--controller-ros-args', '-r /joint_states:=/drive/joint_states',
+                               '--controller-ros-args', '-r /dynamic_joint_states:=/drive/dynamic_joint_states'],
                     ros_arguments=['--log-level', log_level],
                 ),
                 Node(
@@ -94,8 +96,6 @@ def launch_setup(context, *args, **kwargs):
                     parameters=[params],
                     remappings=[
                         ('/controller_manager/robot_description', '/robot_description'),
-                        ('/joint_states', '/drive/joint_states'),
-                        ('/dynamic_joint_states', '/drive/dynamic_joint_states'),
                     ],
                     ros_arguments=['--log-level', log_level],
                 ),
--- a/src/ros/rover/drive/drive_bringup/launch/drive.launch.py
+++ b/src/ros/rover/drive/drive_bringup/launch/drive.launch.py
@@ -94,6 +94,8 @@ def launch_setup(context, *args, **kwargs):
                     parameters=[params],
                     remappings=[
                         ('/controller_manager/robot_description', '/robot_description'),
+                        ('/joint_states', '/drive/joint_states'),
+                        ('/dynamic_joint_states', '/drive/dynamic_joint_states'),
                     ],
                     ros_arguments=['--log-level', log_level],
                 ),
--- a/src/ros/rover/drive/drive_bringup/launch/drive.launch.py
+++ b/src/ros/rover/drive/drive_bringup/launch/drive.launch.py
@@ -86,8 +86,7 @@ def launch_setup(context, *args, **kwargs):
                     package='controller_manager',
                     executable='spawner',
                     arguments=['joint_state_broadcaster',
-                               '--controller-ros-args', '-r /joint_states:=/drive/joint_states',
-                               '--controller-ros-args', '-r /dynamic_joint_states:=/drive/dynamic_joint_states'],
+                               '--controller-ros-args', '-r __ns:=/drive'],
                     ros_arguments=['--log-level', log_level],
                 ),
                 Node(
--- a/src/ros/rover/drive/teleop_drive_joy/launch/teleop.launch.py
+++ b/src/ros/rover/drive/teleop_drive_joy/launch/teleop.launch.py
@@ -19,14 +19,11 @@ def launch_setup(context, *args, **kwargs):
             package='joy',
             executable='game_controller_node',
             name='game_controller_node',
+            namespace='/drive',
             parameters=[
                 {'device_id': device_id,
                  'deadzone': 0.1,
                  'autorepeat_rate': 20.0} | {'device_name': device_name} if device_name else {}
-            ],
-            remappings=[
-                ('/joy', '/drive/joy'),
-                ('/joy/set_feedback', '/drive/joy/set_feedback'),
             ],
         ),
         Node(
--- a/src/ros/rover/drive/teleop_drive_joy/launch/teleop.launch.py
+++ b/src/ros/rover/drive/teleop_drive_joy/launch/teleop.launch.py
@@ -1,8 +1,9 @@
 from launch import LaunchDescription
 from launch.actions import OpaqueFunction, DeclareLaunchArgument
-from launch.substitutions import LaunchConfiguration
+from launch.substitutions import LaunchConfiguration, PythonExpression, PathJoinSubstitution
 from launch_ros.actions import Node
-
+from launch_ros.substitutions import FindPackageShare
+from os.path import expanduser
 
 def launch_setup(context, *args, **kwargs):
 
@@ -10,6 +11,7 @@ def launch_setup(context, *args, **kwargs):
     device_id = LaunchConfiguration('device_id')
     device_name = LaunchConfiguration('device_name')
     joy_vel = LaunchConfiguration('joy_vel')
+    teleop_params = LaunchConfiguration('teleop_params')
 
     return [
         # Add Nodes
@@ -22,11 +24,16 @@ def launch_setup(context, *args, **kwargs):
                  'deadzone': 0.1,
                  'autorepeat_rate': 20.0} | {'device_name': device_name} if device_name else {}
             ],
+            remappings=[
+                ('/joy', '/drive/joy'),
+                ('/joy/set_feedback', '/drive/joy/set_feedback'),
+            ],
         ),
         Node(
             package='teleop_drive_joy',
             executable='teleop_drive_joy_node',
             name='teleop_drive_joy_node',
+            parameters=[teleop_params],
             remappings=[
                 ('/cmd_vel', joy_vel),
             ],
@@ -34,7 +41,23 @@ def launch_setup(context, *args, **kwargs):
     ]
 
 def generate_launch_description():
+    teleop_drive_dir = PythonExpression([
+        '"', PathJoinSubstitution([expanduser("~") + '/nova/src/ros/rover/drive/teleop_drive_joy']),
+        '" if "', LaunchConfiguration('local'), '".lower() == "true" else "',
+        FindPackageShare('teleop_drive_joy'), '"'
+    ])
+
     declared_arguments = [
+        DeclareLaunchArgument(
+            name='local',
+            default_value='False',
+            description='Whether to use the local teleop_drive_joy source directory instead of the nix store for param files.',
+        ),
+        DeclareLaunchArgument(
+            name='teleop_params',
+            default_value=PathJoinSubstitution([teleop_drive_dir, 'params', 'teleop.yaml']),
+            description='The main parameter file to use for teleop_drive_joy_node',
+        ),
         DeclareLaunchArgument(
             name='device_id',
             default_value='0',
--- a/src/ros/rover/drive/teleop_drive_joy/params/teleop.yaml
+++ b/src/ros/rover/drive/teleop_drive_joy/params/teleop.yaml
@@ -1,0 +1,52 @@
+# see teleop_drive_joy.yaml for descriptions given to options
+teleop_drive_joy_node:
+  ros__parameters:
+    initial_speed: 0.1
+    max_input_threshold: 0.97
+    handbrake_speed_multiplier: 0.6
+    trigger_pressed_threshold: 0.1
+
+    rumble_delay: 0
+    rumble_joints: [ frw, flw, brw, blw ]
+    rumble_range: [ 0.4, 0.8 ]
+
+    speed_limit_max: 1.0
+    speed_limit_min: 0.05
+    speed_change_coarse_val: 0.1
+    speed_change_fine_val: 0.025
+
+    input_topic: /drive/joy
+    output_topic: /cmd_vel
+    drive_info_topic: /drive/drive_info
+    joint_states_topic: /drive/joint_states
+    joy_feedback_topic: /drive/joy/set_feedback
+
+    controllers:
+      - pivot_drive_controller
+      - ackermann_steering_controller
+      - strafe_drive_controller
+      - diff_drive_controller
+    pivot_drive_controller:
+      limit_angular: 0.999
+      limit_linear: 1.0
+      scale_angular: 1.0
+      scale_linear: 1.0
+    ackermann_steering_controller:
+      limit_angular: 1.0
+      limit_linear: 1.0
+      scale_angular: 1.0
+      scale_linear: 1.0
+    strafe_drive_controller:
+      limit_angular: 1.0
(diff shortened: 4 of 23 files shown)


src/ros/nova-gui/nova-gui/src/components/drive/DriveModeWidget/DriveModeWidget.tsx
```diff
@@ -67,7 +67,7 @@ const DriveModeWidget: React.FC<IDriveModeWidgetProps> = (
       </CardHeader>
       <CardBody className="flex justify-center flex-col content-center">
         <div>
-          <div className="grid grid-flow-col gap-3 auto-cols-fr">
+          <div className="grid grid-flow-row grid-cols-[repeat(auto-fit,_minmax(10em,_1fr))] gap-3">
```
FelicityMatt, line 70: I don't love that we will probably have 2 rows, when the cameras expand it will push the speed down and likely off the page. Did it look ok when you tested for the ec thing? and rows of 3 and 1 is kinda ugly but it's fine lol. not worth the hassle of idk trying to make it 2-2.

src/ros/rover/drive/teleop_drive_joy/params/teleop.yaml
```diff
@@ -0,0 +1,55 @@
...
+    max_input_threshold: 0.97
+    handbrake_speed_multiplier: 0.6
+    trigger_pressed_threshold: 0.1
+
+    rumble_enable: true
+    # rumble is currently updated every 50ms (period that drive joint states is published at),
+    # so rumble_delay is actually ceiling(rumble_delay / 50) * 50
+    rumble_delay: 110
+    rumble_joints: [ frw, flw, brw, blw ]
+    rumble_range: [ 0.075, 0.2 ]
```
FelicityMatt, line 15: can we set the rumble intensity anywhere? I just had a play around and it can get quite intense so would be good to play around with values as we go through comp sims. 

Also can you link this param file maybe after merging and just be like software psa if you want to adjust the rumble look here.

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.cpp
```diff
@@ -373,6 +407,95 @@ void TeleopDriveJoy::print_controls()
...
+        RCLCPP_ERROR_STREAM(this->get_logger(), C_INFO << "Gamepad disconnected" << C_END);
+      }
+    }
+}
+
+void TeleopDriveJoy::send_drive_info()
+{
+  drive_interfaces::msg::DriveInfo msg {};
+
+  msg.autonomous_mode = autonomous_mode_;
```
FelicityMatt, line 432: hm we don't have gui components for auto or handbrake status, would be good to add but we should get this merged first, maybe if you could create a github issue for it?

src/ros/rover/drive/teleop_drive_joy/params/teleop.yaml
FelicityMatt: looking below I think this is the param, can you maybe add a comment above it just saying "range of rumble intensity" or something.

My follow up is then where can we configure when the rumble is triggered, like what if it want to change it so that it triggers with more effort.

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.yaml
```diff
@@ -4,12 +4,44 @@
 teleop_drive_joy:
   input_topic:
     type: string
-    default_value: "/joy"
+    default_value: "/drive/joy"
```
FelicityMatt, line 7: ty

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.yaml
```diff
@@ -4,12 +4,44 @@
...
+    type: bool
+    default_value: true
+    description: "Enable controller rumbling if wheel effort is sufficiently high"
+  rumble_range:
+    type: double_array
+    default_value: [0.4, 0.8]
+    description: >
+      Game pad rumble linearly increases from none to max as wheel effort increase in range. 
+      If less than range, game pad will not rumble. 
+      If greater than range, game pad will have maximum rumble.
```
FelicityMatt: Game pad rumble **intensity** linearly increases from none to max **of rumble range** as wheel effort increase in range. 
If **effort is** less than **rumble** range, game pad will not rumble. 
If **effort is**  greater than **rumble** range, game pad will have maximum rumble.
**Effort range is [0,1]**

FelicityMatt: some suggestions cause the description confused me, feel free to adjust.

src/ros/rover/drive/teleop_drive_joy/params/teleop.yaml
FelicityMatt: I was like you know what you should just link to the descriptions but you already do at top, should be fine I think 💀

src/ros/rover/rover_description/banksia/urdf/ros2_control.xacro
```diff
@@ -42,6 +42,10 @@
                 <state_interface name="position"> 
                     <param name="initial_value">0.0</param>
                 </state_interface>
+
+                <state_interface name="effort">
```
FelicityMatt, line 46: and just confirming blcmd hardware is already configured to fill this state interfaces?

FelicityMatt (commented):
LGTM

some minor comments, but happy to merge as is given @tytia gives it a quick look, and that it's been tested on rover.

src/ros/rover/drive/drive_bringup/launch/drive.launch.py
```diff
@@ -85,7 +85,9 @@ def spawner_args(controller: str) -> list[str]:
                 Node(
                     package='controller_manager',
                     executable='spawner',
-                    arguments=['joint_state_broadcaster'],
+                    arguments=['joint_state_broadcaster',
```
tytia, line 88: use namespace over remappings when possible, i.e. `namespace = '/drive'`

src/ros/rover/drive/teleop_drive_joy/launch/teleop.launch.py
```diff
@@ -22,19 +24,40 @@ def launch_setup(context, *args, **kwargs):
                  'deadzone': 0.1,
                  'autorepeat_rate': 20.0} | {'device_name': device_name} if device_name else {}
             ],
+            remappings=[
```
tytia: use namespace over remappings when possible

src/ros/rover/drive/teleop_drive_joy/launch/teleop.launch.py
```diff
@@ -22,19 +24,40 @@ def launch_setup(context, *args, **kwargs):
...
+        '" if "', LaunchConfiguration('local'), '".lower() == "true" else "',
+        FindPackageShare('teleop_drive_joy'), '"'
+    ])
+
     declared_arguments = [
+        DeclareLaunchArgument(
+            name='local',
+            default_value='False',
+            description='Whether to use the local teleop_drive_joy source directory instead of the nix store for param files.',
+        ),
```
tytia, line 52: Is the reasoning for this to use `local:=true` instead of having to pass the absolute path of a param file like `teleop_params:=path`?

IMO the PythonExpression is not very readable, and the fact that it has to use a LaunchConfiguration kind of breaks our launch file conventions as we only have LaunchConfigurations in `launch_setup()`

src/ros/rover/drive/teleop_drive_joy/params/teleop.yaml
```diff
@@ -0,0 +1,55 @@
...
+    joint_states_topic: /drive/joint_states
+    joy_feedback_topic: /drive/joy/set_feedback
+
+    controllers:
+      - pivot_drive_controller
+      - ackermann_steering_controller
+      - strafe_drive_controller
+      - diff_drive_controller
+    pivot_drive_controller:
+      limit_angular: 0.999
```
tytia, line 36: can you add a comment here explaining why this is set to 0.999 (encoders limit the pivots to +-90 degrees of rotation, there is a miniscule difference between the offset defined in our hardware interfaces vs. the actual offset defined in firmware)

tytia (commented):
Wonderful work overall, left a few comments

src/ros/rover/rover_description/banksia/urdf/ros2_control.xacro
Glass2721: yes, blcmdhardware is really strange in that it detects if state interfaces exist in the urdf, and then configures itself to use them if so.

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.cpp
Glass2721: auto status is problematic because there is no good way to know if drive is in auto or manual mode before the first press of the set auto mode or set manual mode buttons on the game pad. So I'm inclined to not try to do this, as it would be alot of work to sync that up on startup.

Handbrake however, could be done. I originally was trying to get another progress bar to be overlaid over the current drive speed display in the gui, which could show the "actual drive speed", which is current speed * handbrake speed reduction (if handbrake is pressed). wdyt?

src/ros/rover/drive/teleop_drive_joy/params/teleop.yaml
Glass2721: I suspect there may be some confusion as to the relationship between rumble intensity and wheel current. With `    rumble_range: [ 0.075, 0.2 ]`, the relationship looks like this:

Screenshot From 2026-02-21 19-08-43

where the x-axis is max wheel effort and the y-axis is rumble intensity (both are in [0, 1]).

so in effect, you can configure max rumble intensity by have the upper value of rumble_range be >1, and that the minimum effort required to start rumble is the lower value of rumble range.

I do think there may be merit to capping rumble intensity at a value other than 1, but I didn't see the need for it after testing these settings at ec field test, so probably add later if required (this is due to the fact that current is very spiky, and doesn't stay high for long). wdyt?

> Also can you link this param file maybe after merging and just be like software psa if you want to adjust the rumble look here.

I'll chuck a psa in #drive and notify all the drive ops so they know this is happening, and how to change settings

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.yaml
Glass2721: looks good, except I would correct your first sentence with:
>Game pad rumble **intensity** linearly increases from none to max as wheel effort increases in **rumble** range.

imo, there is no good way of explaining this without writing a thesis #thecodeisthedocumentation

src/ros/nova-gui/nova-gui/src/components/drive/DriveModeWidget/DriveModeWidget.tsx
Glass2721: I don't recall speed being pushed off the page, so not sure if that's going to be a problem. If we wanted to keep it in one row, I'd have to extend the widget over two columns on camera pages, rather than one. wdyt?

> Did it look ok when you tested for the ec thing? and rows of 3 and 1 is kinda ugly but it's fine lol. not worth the hassle of idk trying to make it 2-2.

turns out, for the spaces we insert this widget into, the dimensions always end up being either 1x4 or 2x2, which is nice. I genuinely don't know how to make it always like that though, so maybe help?

src/ros/rover/drive/teleop_drive_joy/launch/teleop.launch.py
Glass2721: the `local` launch argument has sort of become a standard amongst launch files that pass parameter files into launched nodes, and allows us to really quickly make changes to params, relaunch with `local:=true`, and have the changes apply without rebuilding (no need to search for param file path every time).

The way `teleop_drive_dir` is resolved is directly copied from the (many) other launch files we have that do this exact same thing. I agree it sucks, but that's the way we currently do it (unless we want to remove `local` entirely from this launch file).

wdyt?

src/ros/rover/drive/drive_bringup/launch/drive.launch.py
Glass2721: just checking, do we need transforms related to joints measured by drive? because namespacing `/joint_states`to `/drive/joint_states` means that robot state publisher won't get these joint states without some other configuration changes.

src/ros/rover/drive/teleop_drive_joy/launch/teleop.launch.py
Glass2721: maybe change the standard way we do this? idk seems like a lot of work lol

src/ros/rover/drive/teleop_drive_joy/params/teleop.yaml
Glass2721: I can add a comment but something interesting to note: I suspect the issue is not in firmware, but in blcmdhardware2. Because, if you send it the max position, it'll wrap it around to the min position because of this strange calculation:
`auto one_clamped_scalar = std::fmod((value/max) + 3, 2) - 1;`

I suspect if you check can when this issue is occurring, the messages we're sending will accurately reflect the current positions of the pivots.

src/ros/rover/drive/drive_bringup/launch/drive.launch.py
tytia: Yeah that's true, I suggested the change because I assumed that's what your remappings were doing anyway. Actually tbh why was there a need to remap to `/drive/joint_states` and `/drive/dynamic_joint_states`?

src/ros/rover/drive/teleop_drive_joy/launch/teleop.launch.py
tytia: we can keep this for now, we can change it to be nicer later

src/ros/rover/drive/teleop_drive_joy/params/teleop.yaml
tytia: Oohh interesting, might have to look into blcmdhardware2 in that case then

src/ros/rover/drive/drive_bringup/launch/drive.launch.py
Glass2721: I do it because arm does it, and due to the fact that we handle the urdf in an interesting way. Instead of loading the whole urdf into robot state publisher, different launch files load different parts into their own instance of robot state publisher. This causes error messages to appear when e.g. arm and drive are launched together, as joints being broadcast in the urdf loaded by drive aren't in the urdf loaded by arm.

imo this is a real mess and we'll need to clean this up, but probably not in this pr. I believe auto does their own thing with the urdf as well.

src/ros/nova-gui/nova-gui/src/components/drive/DriveModeWidget/DriveModeWidget.tsx
FelicityMatt: hm strange it was 1 and 3 for me, should be fine tho not worth wasting time on.

src/ros/rover/drive/teleop_drive_joy/params/teleop.yaml
FelicityMatt: ah yep sounds good, I get it now :)

src/ros/rover/drive/teleop_drive_joy/src/teleop_drive_joy.cpp
FelicityMatt: mm I like that idea, I think we should get this PR merged in first before trying other things tho.

FelicityMatt (commented):
LGTM

PR #898, opened 2026-02-12
fix: can launch reliability

Glass2721:
## Description

Errors when attempting to log the can bus are now output in red text, and can-utils is now included in nova-workspace so (according to orlando) candump when executed from can.launch.py should use the included candump and not the system one (unreliable).

## Steps to Test

ros2 launch nova_bringup can.launch.py bus:=can0 bitrate:=250000 (for extra testing, you could remove can-utils from your system first)

### DIFF
--- a/src/ros/rover/nova_bringup/launch/can.launch.py
+++ b/src/ros/rover/nova_bringup/launch/can.launch.py
@@ -49,8 +49,8 @@ def start_can(bus: str, bitrate: str, logger: Logger):
         except subprocess.CalledProcessError as e:
             logger.error(f"{Colour.FAIL}Failed to start {bus} with error:{Colour.END}\n {e}")
     else:
-        return logger.warning(f"{bus} is already running "
-                              f"(check bitrate matches requested: {bitrate})")
+        logger.warning(f"{bus} is already running "
+                       f"(check bitrate matches requested: {bitrate})")
 
 def launch_setup(context, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


src/ros/rover/nova_bringup/launch/can.launch.py
```diff
@@ -44,7 +47,7 @@ def can_is_up():
                            capture_output=True, text=True, check=True)
             logger.info(f"{bus} started successfully")
         except subprocess.CalledProcessError as e:
-            logger.error(f"Failed to start {bus}: {e}")
+            logger.error(f"{Colour.FAIL}Failed to start {bus} with error:{Colour.END}\n {e}")
     else:
         return logger.warning(f"{bus} is already running "
                               f"(check bitrate matches requested: {bitrate})")
```
Redecorating: i have done a little research and you can actually check this in code:

`ip -j -d link show can0  | jq .[0].linkinfo.info_data.bittiming.bitrate`

doesn't need to be part of this pr (might be better as part of the can script).

Redecorating (approved):
nice

src/ros/rover/nova_bringup/launch/can.launch.py
Glass2721: cool! making this part of the can script would be nice; though for this launch file, I wouldn't want it to actually change the bitrate (but warning that it is actually different from requested would be great) as that could be intentional and mess stuff up if changed.

PR #899, opened 2026-02-12
Fix/nix build gui

Redecorating:
## Description

- make the gui built by nix build right (before most of the css was missing)
- remove a stray
- add script for serving the gui: result/bin/serve-gui -p 8080 (this will be in ws-build and `nova-build -A pkgs.ros.nova-gui`)
- make a `nova.services.nova-gui.enable = true;` you can put in your configuration.nix to auto serve gui+rosbridge (may be good for rover at outreach events so you can see gui with phone)
- change to mbtileserver which is packaged for nix already, update the alias to `tileserver`. please put your mbtiles in ~/tiles/. you don't need to specify the exact filename as an argument to `tileserver`.
- launch scripts: update gui and tileserver launch to use the built gui and mbtileserver. you should be able to just do `~/Builds/master/launch/run-gui-maps` once this is merged :)

## Memes
image


Redecorating:
btw if we normally put tile files in ~/maps please change what the alias I added here uses to whatever we already have been using as the folder for the tiles

FelicityMatt:
yeah we've been using ~/maps as the folder

FelicityMatt (approved):

PR #902, opened 2026-02-14
Adding nav2-route (updating nix-ros-overlay)

### LINKED ISSUE #906: Adding nav2-route
### **Context**

Nav2 route server is a package we want to use for auto ARCh 2026. However, it's unavailable in our current pinned revision of nix-ros-overlay. 
Originally, updating nix-ros-overlay to the latest version of nav2-route required updating nixpkgs, which deleted freeimage, which is a dependency for Gazebo, breaking it (there's an [open PR](https://github.com/lopsided98/nix-ros-overlay/pull/746) on nix-ros-overlay at the moment to fix this, but it's not finished).
Currently, instead of trying to updating to the latest version of nav2-route (1.3.11), we'll update to an earlier version (1.3.9) where nix-ros-overlay used an older version of nixpkgs than the one we currently have pinned, meaning we wouldn't need to update nixpkgs.

### **TODO**

- [x] Update nix-ros-overlay to dd31a43
- [ ] Add nav2-route

Tags for Notion Integration:
tag:autonomous;tag:nixfiles

vic-bart:
## Memes

Tags for Notion Integration:
tag:autonomous;tag:software;tag:nixfiles


vic-bart:
I am concerned about my fix to rtabmap, but if it works then there's probably no reason to worry about it right this moment and the PR can be merged in. To be tested.

PR #907, opened 2026-02-16
Feat/nixfiles/rpi4

Redecorating:
## Description

nix configuration for raspberry pi 4

also make a smaller variant of the workspace with just the stuff the raspberry pi needs.


nixfiles/modules/nixos/common/substituters/default.nix
rotaryViper, line 1: Maybe don't rely on ../ in case the directory changes

Redecorating: @vic-bart are you gonna change it again?

Redecorating: (the above is a joke). We are already doing it for the publicKey just below so it should be fine

PR #908, opened 2026-02-17
[WIP] Added configs for new devices

rotaryViper:
## Description

It worked on my own laptop.
If you have an intel 12th or newer laptop, run:

devices.laptop.aftershock-light.enable = true

In your configuration.nix

To do:
- Find out which nvidia gpu is on older laptops
- Find out the bus id for nvidia gpu and intel igpu to get prime working for older laptops
- Get everyone's opinion on the new performance.nix file

## Changelogs

Added the following options for devices.laptop:
- enable
Enable all hardware and firmware
Enable all detected operating systems on boot, and 30s boot menu
Enable all unfree software

- amd.enable
Enable amd cpu latest microcode
Enable amd cpu to respond to compute loads faster.
Enable amd igpu early in boot to fix boot resolution.
Enable amd igpu opencl.

- intel.enable
Enable intel cpu latest microcode
Enable intel cpu to read balance cpu performance over time with more temperature sensors
Enable intel igpu acceleration libraries for opencl, compute and media
intel-new for intel hybrid cpus | intel-old for before hybrid cpus

- nvidia.enable
Enable nvidia gpu latest drivers
Enable gpu offloading to save power when nvidia gpu is not necessary, while letting nvidia gpu take heavy loads
Enable nvidia caches
Enable cuda if gpu allows

- performance.enable
Swap i/o scheduler to remove the algorithm for fast flash storage that doesn't need any algorithm. Saves power and increases performance for most of the team.
Enables ssd maintainence service.
Use a userspace cpu scheduler that keeps your laptop responsive, even under heavy load such as compiling code.
Depriortize the nix daemon, so that user actions take priority over nix commands. Sacrifices some nix build speed to desktop responsive when nix-building.
De-duplicates nix store once a week to save space without losing any data.

## Steps to Test

device.laptop.performance.enable for performance optimizations.
Try device.laptop.metabox-new.enable on the shiny metabox


FelicityMatt:
@rotaryViper thanks for getting this working but I'm a bit confused why this is being added to the mono repo for specific laptops? should we not add it to [the nix-configs repo](https://github.com/MonashNovaRover/nix-configs)?? isn't that the entire point of that repo?

nixfiles/modules/nixos/devices/laptop/aftershock-light/default.nix
```diff
@@ -0,0 +1,16 @@
+{ config, lib, ... }:
+
+let
+  cfg = config.devices.laptop.aftershock-light;
+in
+{
+  options.devices.laptop.aftershock-light.enable = lib.mkEnableOption "configuration for the Aftershock Light";
```
Redecorating: this one needs #911  if ur interested in figuring that out at some point

nixfiles/modules/nixos/devices/laptop/default.nix
```diff
@@ -0,0 +1,49 @@
+{ config, lib, ... }:
+
+let
+  cfg = config.devices.laptop;
+in
+{
+  imports = [
+    # Laptops
+    ./aftershock-heavy
```
Redecorating, line 9: i think it may be better to use the model name `cat /sys/class/dmi/id/*name`

nixfiles/modules/nixos/devices/laptop/default.nix
```diff
@@ -0,0 +1,49 @@
...
+    ./performance.nix
+  ];
+
+  options.devices.laptop.enable = lib.mkEnableOption "Nova Rover base station laptop configuration";
+
+  config = lib.mkIf cfg.enable {
+
+    # Enable everything
+    hardware = {
+      enableAllHardware = lib.mkDefault true;
```
Redecorating, line 30: this seems to just add a lot of modules to the initrd, I don't know if we need that cause nixos-generate-config should add whichever specific modules are needed to hardware-configuration.nix. does anything not work without this? I worry that it will make the initrd bigger which means we run out of space in /boot faster

nixfiles/modules/nixos/devices/laptop/default.nix
```diff
@@ -0,0 +1,49 @@
...
+      enableAllFirmware = lib.mkDefault true;
```
Redecorating, line 31: this adds xbox firmware too lol. I don't think this has the same impact as the other one in terms of initrd size so this one is probably ok

nixfiles/modules/nixos/devices/laptop/default.nix
```diff
@@ -0,0 +1,49 @@
...
+    };
+
+    # Detect other OS on boot
+    boot.loader = {
+      efi.canTouchEfiVariables = lib.mkDefault true;
+      systemd-boot.enable = lib.mkDefault true;
+      timeout = lib.mkDefault 30;
```
Redecorating: does it need to be 30s? 99% of the time you don't want to change anything when booting, otherwise you just need a second or two to press an arrow key or smth to stop the timer entirely

nixfiles/modules/nixos/devices/laptop/default.nix
```diff
@@ -0,0 +1,49 @@
...
+    };
+
+    nixpkgs = {
+      # Allow non-free drivers
+      config.allowUnfree = lib.mkDefault true;
+
+      # Define platform as x86_64
+      hostPlatform = lib.mkDefault "x86_64-linux";
```
Redecorating, line 55: @tytia L

Redecorating: u don't have a laptop anymore

nixfiles/modules/nixos/devices/laptop/nvidia.nix
```diff
@@ -0,0 +1,49 @@
...
+      };
+    };
+
+    # Switching between igpu and nvidia gpu
+    services.switcherooControl.enable = true;
+    services.xserver.videoDrivers = lib.mkDefault [ "nvidia" ];
+
+    # Enable nvidia caches
+    nix.settings = {
+      substituters = [
```
Redecorating: this can go to nova/nixfiles/modules/nixos/common/substituters/default.nix maybe? we can move the jetpack copy of this substitutor there too

nixfiles/modules/nixos/devices/laptop/default.nix
Redecorating: although in software meeting they wanted these files moved out to nix-configs so idk

nixfiles/modules/nixos/devices/laptop/default.nix
rotaryViper: General config files cater to the majority, mkDefault has a lower priority than normal settings so should be ok

nixfiles/modules/nixos/devices/laptop/nvidia.nix
rotaryViper: very good sugestion, I'll add both of them there

nixfiles/modules/nixos/devices/laptop/default.nix
rotaryViper: Uh should be ok? I get 55mb per initrd with my setup from enabling all hardawre

nixfiles/modules/nixos/devices/laptop/default.nix
rotaryViper: It was the default setting on the metabox, i iwll change this back to 10 seconds tho which is the default frmo when you first install nixos

rotaryViper:
imma merge it

PR #917, opened 2026-02-22
Feat/kiln door control

### LINKED ISSUE #791: Kiln Door Control
### **Context**

Science has a new door design for their kiln that needs to be controlled. I think the easiest thing to do is buttons on the joysticks to press to move in one direction and maybe using the slider to adjust speed.

This will be effort controlled, through the QCMD.

Additionally they would like to current sense the motor to tell how hard the door is pressed against the kiln, so we will need to collect this data and show it on the GUI somewhere.

### **TODO**

- Write the door controller to take inputs from teleop and move the door.
  - Use the existing CMDHardware interface using effort 
- Update GUI

Tags for Notion Integration:
tag:software;tag:science

angelaguinaldo:
## Description

Adds the kiln door controller to science and a current sensor emulator into can sleuth.

## Steps to Test

1. Start can1 or vcan1
2. `ros2 run science chute.py`
3. `ros2 launch teleop_science teleop.launch.py joystick:=false`
4. Optional: start the current sensor emulator using `nova-build -A pkgs.python3Packages.nova-can-sleuth -o /tmp/can_sleuth`
5. Then run `/tmp/can_sleuth/bin/can_sleuth current_sensor_emulator`
6. Have a play around by can sending values to the door qcmd (with id 0D2) and the current sensor (4FF). You can also change the maximum current value of the door and the current sensor emulator update function.

## Memes
aks0x5

### DIFF
--- a/src/other/can_sleuth/can_sleuth/systems.py
+++ b/src/other/can_sleuth/can_sleuth/systems.py
@@ -13,7 +13,7 @@ EDITED BY: Orlando Chamberlain
 
 from can_sleuth.devices import blcmd
 from can_sleuth.devices import blcmd_emulator
-from can_sleuth.devices import sensor
+from can_sleuth.devices import sensor_emulator
 import random
 
 from can_sleuth.devices import led as LED #gotta rename some things
@@ -87,7 +87,7 @@ def current_sensor(interface="can1", emulate=False):
         return random.randint(-32767, 32767)
 
 
-    currentSensor = sensor.Sensor("current sensor", canId, interface, "current", "amps", 0, update_function)
+    currentSensor = sensor_emulator.SensorEmulator("current sensor_emulator", canId, interface, "current", "amps", 0, update_function)
 
     return [currentSensor]
 
@@ -99,7 +99,7 @@ allSystems = {
         "taipan": taipan_spherical,
         "drive25_26": drive25_26,
         "taipan_spherical": taipan_spherical,
-        "current_sensor": current_sensor,
+        "current_sensor_emulator": current_sensor,
         "led": led
         }
 
--- a/src/other/can_sleuth/can_sleuth/systems.py
+++ b/src/other/can_sleuth/can_sleuth/systems.py
@@ -13,6 +13,9 @@ EDITED BY: Orlando Chamberlain
 
 from can_sleuth.devices import blcmd
 from can_sleuth.devices import blcmd_emulator
+from can_sleuth.devices import sensor
+import random
+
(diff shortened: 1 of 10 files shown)


src/other/can_sleuth/can_sleuth/systems.py
```diff
@@ -73,12 +76,30 @@ def led(interface = "can0", emulate=False):
...
+
+    
+
 # List of everything for help message:
 allSystems = {
         "drive": drive25_26,
         "taipan": taipan_spherical,
         "drive25_26": drive25_26,
         "taipan_spherical": taipan_spherical,
+        "current_sensor_emulator": current_sensor,
```
Redecorating: could you make this "science25_26"/"science" maybe? Then we can add more things to it later for the rest of the payload.

Redecorating (approved):
Nice!

src/other/can_sleuth/can_sleuth/systems.py
angelaguinaldo: yep can do

src/ros/rover/science/science/science/arc/kiln_door.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        super().__init__(contexts)
+        self.logger.info(f"KilnDoor -- I have been __init__ialized")
+
+        #ros2 parameters
+        self.door_actuation_axis_name = self.declare_parameter("actuation_axis", "door_actuation").value
+        self.door_speed_name = self.declare_parameter("speed_axis", "door_speed").value
+
+        #maximum door current 
+        self.max_door_current = self.declare_parameter("max_door_current", max_door_current)
```
FelicityMatt: this shouldn't be in the controller, we already have percent effort in the qcmd hardware, controllers should never concern itself with things like current, that is a hardware interface level thing.

I think you've commented it out below as well, we can just remove that.

src/ros/rover/science/teleop_science/params/arc/joysticks.config.yaml
```diff
@@ -114,10 +114,16 @@ thrustmaster_right:
           from: "r_ax_stick_y"
         auger_drill_speed:
           from: "r_ax_slider"
-        # Chute
+                # Chute
```
FelicityMatt: can we remove this change lol

FelicityMatt (commented):
looking good, some small comments

FelicityMatt (commented):
If you guys have tested this, lgtm

PR #918, opened 2026-02-22
[wip] blcmd hardware2 multiturn position

Redecorating:
## Description

```
ros2 launch arm_bringup control.launch.py local:=true rviz:=true | grep -v -e TF_NAN_INPUT -e "not defined in"
```

```
ros2 launch teleop_arm teleop.launch.py joysticks:=false log_inputs:=True
```

Drive will not load on this branch. Please just test with the arm. J456 are hard coded to use generic mock hardware instead of blcmd_hardware2, so no can messages will be sent to them and telemetry will be ignored.

for J2 and J3, please adjust the "home_offset" in /home/nova/nova/src/ros/rover/rover_description/banksia_arm/urdf/arm.ros2_control.xacro to whatever the right angle (urdf zero) position is in terms of resolver feedback. I think J1 is just 0x2000 so that should be fine.

to test without an actual arm use can_sleuth

```
 ~/Builds/j123/bin/can_sleuth -o tui -e taipan
```

note that the angle isn't correct in the cansleuth emulator (small box with just pos and vel) but the tracer (bigger box) seems correct. this shows the angle of the resolver, not the joint.

need to fix it for drive and add in some functionality i cut out of blcmd-hardware. If someone tells me what angle the drive pivots are zeroed at (either in radians or resolver ticks) that'd be great.

if someone knows what the units are for velocity control in general with blcmds that'd be great too.


Redecorating:
also need to make sure the whole arm doesn't fail to init if one blcmd is missing

Redecorating:
doing this on another branch

PR #921, opened 2026-02-25
Feat/auto-cameras

tytia:
## Description

Set up the RealSense D415 camera with ROS2.

## Changelogs

- added ros2 packages for realsense and usb cameras
- patched librealsense so it can build on nixos
- updated udev rules
- added v4l-utils to shared nixfiles
- added nom and btop to shared nixfiles (unrelated)
- renamed camera.launch.py to oak.launch.py
- added d415 to URDF
- set up launch file
- configured d415 resolution and frame rate to 640x360 at 15 fps, disabled all other streams except colour

## Steps to Test

1. `ros2 launch auto_bringup realsense.launch.py`
2. `ros2 launch auto_bringup urdf.launch.py joints:=true`
3. `launch-rviz`

To see the d415 model in rviz, go to RobotModel -> links and disable `oak_base_frame`

## Memes

image


vic-bart:
LGTM, but I'm concerned about the accuracy of the additions to the URDF; if those poses are inaccurate, it will reduce the accuracy of the stack, so they should be within +/- <5 mm at least.

PR #922, opened 2026-02-25
blcmd hardware multiturn position

Redecorating:
## Description

need to check if this is correct for drive - i did that on my other branch but should reverify it still


jacksonmacc (approved):
looks good to me!

PR #925, opened 2026-02-26
Ec launch scripts

TJ-Lee-M:
## Steps to Test
start vcan0
run`~/Builds/result/launch/ec-one localhost` or ec-two or ec-combined  (or rover ip)

### DIFF
--- a/nixfiles/modules/home/macros/launch/arch/excavation-construction.nix
+++ b/nixfiles/modules/home/macros/launch/arch/excavation-construction.nix
@@ -11,7 +11,9 @@ let
   one = {
     pre = pre-shell {payload-name=task-name + " one"; need-rover=true;};
     terminals = [
-      {name = "Base:Rviz"; platform=base; cmd="ros2 launch auto_bringup rviz.launch.py";}
+      {name = "Base:ecTeleop"; platform=base; cmd="ros2 launch teleop_ec teleop.launch.py log_inputs:=true";}
+      {name = "Rover:EC"; platform=rover; cmd="ros2 launch nova_bringup ec_rover.launch.py";}
+      {name = "Rover:Cameras"; platform=rover; cmd="~/Builds/cameras2legacy/bin/gst-nova-launcher ros2 launch cameras2 camera_server_launch.py platform:=orin param-dir:='\''/home/nova/nova/src/ros/cameras2/cameras2/params'\'' payload:=ec";}
     ];
     post = post-shell;
   };
@@ -19,8 +21,10 @@ let
   two = {
     pre = pre-shell {payload-name=task-name + " two"; need-rover=true; };
     terminals = [
-      {name="Base:Teleop"; platform=base; cmd="ros2 launch teleop_drive_joy teleop.launch.py";}
-      {name = "Rover:Drive"; platform=rover; cmd="launch-drive";}
+      {name="Base:driveTeleop"; platform=base; cmd="ros2 launch teleop_drive_joy teleop.launch.py";}
+      {name = "Rover:Drive"; platform=rover; cmd="ros2 launch drive_bringup drive.launch.py";}
+      {name="Base:Reolink"; platform=base; cmd="reolink"}
+      {name="Base:Reolink"; platform=base; cmd="reolink-ctl"}
     ];
     post = post-shell;
   };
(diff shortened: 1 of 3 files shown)


nixfiles/modules/home/macros/launch/arch/excavation-construction.nix
Glass2721, line 1: change `ros2` to `./ros2` (run from current directory, not PATH)

nixfiles/modules/home/macros/launch/arch/excavation-construction.nix
```diff
@@ -11,16 +11,19 @@ let
   one = {
     pre = pre-shell {payload-name=task-name + " one"; need-rover=true;};
     terminals = [
-      {name = "Base:Rviz"; platform=base; cmd="ros2 launch auto_bringup rviz.launch.py";}
+      {name = "Base:ecTeleop"; platform=base; cmd="ros2 launch teleop_ec teleop.launch.py log_inputs:=true";}
+      {name = "Rover:EC"; platform=rover; cmd="ros2 launch nova_bringup ec_rover.launch.py";}
+      {name = "Rover:Cameras"; platform=rover; cmd="~/Builds/cameras2legacy/bin/gst-nova-launcher ros2 launch cameras2 camera_server_launch.py platform:=orin param-dir:='\''/home/nova/nova/src/ros/cameras2/cameras2/params'\'' payload:=ec";}
```
Glass2721: I think you can just copy the `cameras-orin` alias definition directly from `/home/nova/nova/nixfiles/modules/home/macros/default.nix`, this probably works but is a bit strange with the escaping around param-dir

nixfiles/modules/home/macros/launch/arch/excavation-construction.nix
Glass2721, line 1: add reolink-ctl and gui to terminals launched?

Glass2721 (changes requested):
looks good, few comments

nixfiles/modules/home/macros/launch/arch/excavation-construction.nix
TJ-Lee-M: yeah we can just use alias, ive just got the full cmd for now bc we have to change btwn master & cameras2legacy depending on the platform atm bc not all are built the same. once thats fixed I agree alias is better

nixfiles/modules/home/macros/launch/arch/excavation-construction.nix
TJ-Lee-M: i left gui and rosbridge out bc we can launch them earlier (before going into base station & before connecting to rover). we can try including them and doing all at once but I think keeping seperate will just save a little time

reolink-ctl I left out bc I usually have it in another window for easier access but im sure we can automate that launching too - ill have a look at it. (also the alias is missing from recently built machines atm rip)

nixfiles/modules/home/macros/launch/arch/excavation-construction.nix
```diff
@@ -11,16 +13,20 @@ let
...
 
   two = {
     pre = pre-shell {payload-name=task-name + " two"; need-rover=true; };
     terminals = [
-      {name="Base:Teleop"; platform=base; cmd="ros2 launch teleop_drive_joy teleop.launch.py";}
-      {name = "Rover:Drive"; platform=rover; cmd="launch-drive";}
+      {name="Base:driveTeleop"; platform=base; cmd="./ros2 launch teleop_drive_joy teleop.launch.py";}
+      {name = "Rover:Drive"; platform=rover; cmd="./ros2 launch drive_bringup drive.launch.py";}
+      {name="Base:Reolink"; platform=base-nix "nix-shell -p ffmpeg"; cmd="reolink";}
+      {name="Base:Reolink-ctl"; platform=base-window-nix "nix-shell -p ffmpeg"; cmd="reolink";}
```
Glass2721: change `reolink` to `./reolink-ctl` (use relative path) and I don't think we need ffmpeg for this so probably `platform=base` is good enough

Glass2721 (changes requested):
almost there!

nixfiles/modules/home/macros/launch/arch/excavation-construction.nix
TJ-Lee-M: good catch thanks :)

TJ-Lee-M: fixed now (used full build path for reolink-ctl bc of the missing alias recently - I thing merge like this to practise with and update with fixed alias later)

Glass2721 (approved):
lgtm! merge after you've taken a brief look and are happy with what `ec-one` does

PR #926, opened 2026-02-27
feat: update science CAN IDs and servos

binudakal:
## Description

Changes chute servo from positional to continuous, adds gear ratio to positional servo hardware for tool rotator, updates all science CAN IDs to CAN standards (except sensors), tested on payload yesterday.


FelicityMatt (approved):
lgtm

PR #929, opened 2026-02-28
feat: Analysis Arm position control

FelicityMatt:
## Description

Position control for the analysis arm and updating tof sensor with new can standards.

Can now be controlled by either the gui with position control, or the joystick when activated.
Position control will override effort control, where you can only use effort when no position has been set.

## Changelogs

- Added a new GUI analysis arm widget.
- Added a number of new services.
- `TriggerHardware` for when you just need to send one can command
- `EventCollection` in contexts, allows you to emulate service calls / trigger events between hardware and controllers (since all on the same process)
- Updated TOF sensor to use new can standards
- Updated `StepperHardware` to keep track of position.
- Updated `AnalysisArmController` to use position control.

## Screenshots

[Screencast From 2026-02-28 12-58-22.webm](https://github.com/user-attachments/assets/eecf8280-0213-451f-a8f4-cc63fb1339d2)

## Steps to Test

1. build branch
2. run gui
     - you will need to run rosbridge from this branch since there is a new service interface (and run `gui-link`)
4. run code
    ```sh
    ros2 launch science_bringup arc.launch.py
    ```
4. Test it out on the gui and run a can dump.

## Memes

image


src/ros/rover/science/science/science/analysis_arm_stepper.py
```diff
@@ -30,13 +202,21 @@
...
-        hardware_name="actuation",
-        actuation_axis="analysis_arm_actuation"
-    ) \
-        .with_hardware("actuation", StepperHardware, can_id=0x0E4) \
+            "controller",
+            AnalysisArmController,
+            actuation_axis="analysis_arm_actuation",
+            max_position=270.0
+        ) \
+        .with_hardware("actuation", StepperHardware, can_id=0x0E6,
```
TJ-Lee-M: is 0E6 right for stepper?

src/ros/rover/science/science/science/analysis_arm_stepper.py
```diff
@@ -30,13 +202,21 @@
...
+            max_position=270.0, use_max_position=True,
+            position_to_steps=lambda x: math.floor((x / 0.04) + 0.5), # rounding each time reduces floating point errors.
+            steps_to_position=lambda x: round(x * 0.04, 2),
+        ) \
         .with_teleop(inputs) \
-        .with_activation_buttons(start_active=True, active_button_name="na", inactive_button_pool_names=["na"]) \
+        .with_activation_buttons(
+            start_active=True,
+            active_button_name="activate_analysis_arm",
```
TJ-Lee-M, line 218: having trouble getting analysis arm inputs to candump... also reactivating after deactivting for cbeam

going to class now, i can try looking again later

src/ros/rover/science/science/science/analysis_arm_stepper.py
FelicityMatt: tis not I will update

image

FelicityMatt: updated when rebasing

src/ros/rover/science/science/science/analysis_arm_stepper.py
FelicityMatt: activating and deactivating is working fine for me (note it won't deactive the gui input, only teleop). - make sure you're using teleop science param from this branch?

But I am investigating the lack of teleop input that is a known issue.

FelicityMatt: wait i'm an idiot

give me a sec

FelicityMatt: should be fixed and working now :+1:

binudakal:
amazing work, tested with updated byte order and works!!

edit: the TOFHeight widget errors when out of range, but not for sensor errors sent in the can frame, is this worth doing here or separate pr?

Nova-Jetson:
@binudakal wdym errors out? - but might fix seperately since I need this in

binudakal:
> @binudakal wdym errors out? - but might fix seperately since I need this in

the widget correctly tells you if the distance reading is out of range, but not if the sensor itself has errored in some way through the error data sent in the can frame, this may be something worth adding later

FelicityMatt:
ah yeah the widget hasn't been updated at all, it's still whatever it was two years ago, so yeah probs worth updating but idk if we're gonna even use it?

PR #930, opened 2026-02-28
nixfiles: don't have two copies of can utils

Redecorating:
consolidate some extra packages in modules/home/nova instead of having them spread out.

make the workspace enabled by default with the shared profile.


Glass2721:
feel free to do whatever changes you need to fix this issue, I'm kinda lost to reproducing it and the fix in general. Also, maybe look into `FindPackagePrefix` to find the absolute path to candump?

Glass2721 (approved):
lgtm, idk if there's any better way to get the path to candump (maybe cmake?)

PR #932, opened 2026-02-28
Feat/configure-auto

tytia:
## Description

This branch was originally made to configure MPPI params, but has been expanded to include some other changes.

## Changelogs
- ### MPPI
  - Configured a 'regular' and a 'slow' version which have maximum forward velocities of 0.3 and 0.1 m/s respectively
  - Use the `mppi_params` launch argument in `navigation.launch.py` to override with `regular.yaml` or `slow.yaml`, otherwise use the params configured in `controller_server.yaml`, `local_costmap.yaml` and `velocity_smoother.yaml`
  - Left comments in relevant locations to explain how tweaking particular params affect MPPI's behaviour (`controller_server.yaml`, `local_costmap.yaml` and `velocity_smoother.yaml`)
  - Enabled optimal trajectory visualisation
    - Applied a patch to make this efficient
- ### Other changes
  - Increased obstacle detection range of voxel layer to 5m and rtabmap obstacles detection to treat slopes flatter than 45 degrees as ground
  - Updated rviz config to include fastlivo2 mapping visualisation (off by default) and use "Best Effort" QOS for realsense image topic
  - Cleaned up ARCh behaviour trees, we now use `nav_to_pose.xml` and `nav_through_poses.xml` under `behavior_tree/arch`
  - Added aliases to switch between Fast DDS and Cyclone DDS, `use_fastdds` and `use_cylonedds`
    - The switch will last until reboot, upon which it goes back to the default (Fast DDS)
    - Cyclone DDS is recommended by Nav2, but we have been using Fast DDS the whole time. Ideally we want to try running auto with Cyclone DDS and check for any improvements in network reliability/quality.


vic-bart:
Changes look reasonable. I'm out of the loop on Nav2 dev so I can't speak to it. I noticed the integration of Cyclone DDS, that might be the only source of issues when it comes to affecting other subteams, but if it's been verified to not do that, then this LGTM.

V01DBREAKER:
LGTM provided this doesn't break alias autocomplete

PR #933, opened 2026-03-01
Depreciated nova-workspace

rotaryViper:
## Description

Deleted beraking option in nova.workspace.enable.
It was a legacy gui toggle.

Depreciated ros package autocomplete until we fix the directory for it


Redecorating:
what is the error that you are getting with nova.workspace.enable = true? I'm not getting any error on https://github.com/MonashNovaRover/nova/pull/930/changes which enables it by default.

PR #935, opened 2026-03-02
launch: update launch scripts for pl one

Redecorating:


V01DBREAKER:
what does
| grep -v [not.defined.in](http://not.defined.in/)" do?
[10:22 PM]( line 14)
[10:22 PM]otherwise lgtm
[10:22 PM]as long as it works how u like
[10:22 PM]also u can nix-shell for these scripts too, theres a thing i made for it, for the ffmpeg thing

V01DBREAKER (approved):
line 20 could likely be done with `ssh-nix-terminal` from `nixfiles/modules/home/macros/launch/default.nix` line 29
otherwise LGTM

PR #936, opened 2026-03-02
nixfiles: networking: fix rover to not use bridges

### LINKED ISSUE #928: ros dds dies with linux bridge networking
### **Context**

the ros dds does not work well when there is a linux bridge interface involved (virtual switch). Need to remove it from the nova networking config that the nixos orin uses till we figure out what's the deal.

It will be hard to do redundancy without bridges although maybe netfilter forwarding rules could do it.

### **TODO**

-

Redecorating:
also move can bus naming to the orin nano section


Redecorating:
closes #928

PR #940, opened 2026-03-05
Fix/can sleuth usability

Redecorating:
## Description

- make tui default
- don't die if messages with wrong length are received
- make values dim if they are old
- add aliases and doc in rover-help

## Steps to Test

can start vcan1
./result/bin/can_sleuth -e taipan

# simulate resolver error:
cansend can1 410#0002

it should go dim after a second so you know the error is gone.

# message with unexpected length:
cansend can1 414#00000000

should not crash

### DIFF
--- a/src/other/can_sleuth/can_sleuth/outputs/tui.py
+++ b/src/other/can_sleuth/can_sleuth/outputs/tui.py
@@ -12,6 +12,7 @@ EDITED BY: Orlando Chamberlain, Will Middlewick
 ''' 
 
 import curses
+import time
 
 from . import output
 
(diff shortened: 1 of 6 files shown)


src/other/can_sleuth/can_sleuth/outputs/tui.py
```diff
@@ -90,6 +91,12 @@ def update(self, devices):
                 height = 1
                 encoder_err = False
                 for attr in dev.attrs:
+                    # make attr values that are old dim
```
TJ-Lee-M, line 94: doesnt seem to dim for me?

TJ-Lee-M (commented):
mostly lgtm! timeout/dim doesnt seem to work for me when I simulate the error...

src/other/can_sleuth/can_sleuth/outputs/tui.py
Redecorating: hmm that's weird. i've retested it and it works for me. i did change the timeout for dimming to 0.2 seconds so maybe that's too fast?

Redecorating:
we'll have to see if the dimming works on rover later

PR #942, opened 2026-03-05
feat: Tool Rotator GUI

FelicityMatt:
## Description

Add a gui for the tool rotator to set the preset position as well as to move to the specified position.

> [!NOTE]
> I did not implement twitching on the gui because it required either me to publish position data or for me to create yet another service and service interface and it felt like too much and too complex for something I don't think will be used (we twitch for ilmenite and nir but software will be preoccupied with taking photos/readings so I think twitching should be done by the operator.

## Changelogs

- Added a `ToolRotatorWidget`
- Updated the `PresetTwitchController` to have services that allow setting of position and presets.

## Screenshots

image

## Steps to Test

- Run the gui and the `tool_rotator.py` node

## Memes

image


binudakal:
lgtm :)

PR #943, opened 2026-03-05
fix: make game pad rumble much less sensitive

Glass2721:
## Description

Urgent change to stop game pad from constantly rumbling (which is extremely distracting and makes drive much more difficult to operate).


Redecorating (commented):
LGTM

PR #946, opened 2026-03-08
taipan: limits

Redecorating:
i was able to get it stuck with l2' and l1' colliding and it not letting me move away from that collision but i changed the boxes a bit and didn't get it again so idk if its still an issue.

need to confirm this doesn't impact end effector

blcmd_hardware: ensure pitch commands are same direction as feedback nova_arm_controller: pramaterise if we collect position/velocity feedback
arm_bringup: reenable position and collision limits for nova arm controller
rover_description: taipan: add collider for l2', extend collider for l1' teleop_arm: swap wrist pitch direction as I think the change to blcmd hardware will have swapped it.


Redecorating:
i forgot to say i also changed the diff wrist pitch zero position to directly down

jacksonmacc (approved):
looks good! i ran it in sim for a bit and it worked perfectly

Redecorating:
does the end effector still work?

jacksonmacc:
do you mean the thing where only one finger appears?

Redecorating:
nah i just haven't checked it the end effector still moves with these changes to nova arm controller.

Redecorating:
I have made some more changes to manually do the limiting for the 4 bar linkage as with the collision box limits for that it got stuck a lot. Its a very slop code implementation but there are parameters to tune the limit.

Redecorating:
reminder to me (or you) to turn back on the do_4bar param

PR #950, opened 2026-03-10
feat: drive autolock and drive controls widget

Glass2721:
## Description

Implements "drive autolock" (i.e. automatic locking if too many blcmd errors) and a new ui widget on every cameras page showing drive locked status, a button to override autolock, and gamepad connected status. There are many more changes so please see Changelog for details.

## Changelog

- implement "drive autolock" feature, which causes drive to automatically lock when too many errors are received from blcmds or that they have been continuous for a certain period of time
- update teleop drive joy (i.e. drive teleop) to override autolock when the left trigger is held down (by implementing a new system to handle axis callbacks when axis values are in specified ranges)
- add ui widget "Drive Controls" to every cameras page in arc gui; includes a display on whether drive is locked (and the reason for that locking), whether gamepad is connected or not, and a toggle button to override drive autolock by dynamically changing a parameter in drive teleop (this widget also displays an error toast when drive locks unexpectedly)
- lock controller when gamepad is disconnected, with stop command to drive controllers (this is a separate feature from drive autolock)
- make teleop drive use new params when updated via service call
- updated blcmd status monitor to publish correct messages for drive autolock to work, add rate limiting to printed error/warning/info messages, and add support for all possible error/warning/info/gate drive condition messages from blcmd (to print/publish), including the new ones

## Steps to Test

dm me and we can get on a vc for me to show you how to test this (it's too complicated to write here).


Redecorating (approved):
nice

PR #951, opened 2026-03-11
feat: add more pl launch scripts

jacksonmacc:
## Description

Add more PL launch scripts, for mock and for drive/cameras. Also fixes an issue in the existing one where it needed a shell with ffmpeg.

Hasn't been checked on the actual rover yet, but works with vcan.

## Changelogs

Added new launch scripts:
* pl-mock
* pl-mock-controller
Updated:
* pl-one
* pl-two

## Steps to Test

Run the scripts from build/launch, using the corresponding names (pl-one, pl-two, etc)

## Memes

image


Redecorating (approved):
nice! I have made a couple of changes to my half. if those look ok to you then please merge!

PR #962, opened 2026-03-13
Feat/arm live limit changes

### LINKED ISSUE #959: nova arm controllers params listener test listener
### **Context**

need to be able to edit limits live. I think we need to implement this https://github.com/PickNikRobotics/generate_parameter_library?tab=readme-ov-file#dynamic-parameters

we may need to reinit the JointSaturationLimiter after params change.

then i think `rqt_reconfigure` is a gui you can edit params in.

### **TODO**

-

Redecorating:
## Description
also make arm controller not spam when joystick is locked
update qcmd id (please double check this!)
also make it impossible to enter position control (a side effect of this is that if you restart control, pressing the trigger down and releasing it should trigger teleop to activate the veolcity controller so you shouldn't have to restart teleop as much)

also disable j5 limits as there were issues when I wasn't there? We should reenable it once we figure out what's going on there.

## Screenshots
Screenshot from 2026-03-13 23-39-46

## Steps to Test

use rqt to change limits.

1. open ~/Builds/master/bin/rqt
2. from the top menu bar: plugins -> configuration -> dynamic
   reconfigure
3. on the left, expand the "arm" dropdown
   if it doesn't appear, click the "R" at the top right
4. select nova_arm_velocity_controller

The relevant params are:

do_4bar: do the limit for the 4 bar linkage crunch
j2_minus_j3.{max_position, min_position}: the values for the 4 bar
crunch limit

joint_limits.jX.has_position_limits: Changing this live doesn't work
joint_limits.jX.{min, max}_position: change these to change the limits.
if you want do disable it then pick some large number.

use_limits: global enable disable for joint_limits and the 4bar limit
use_collision_limits: enable disable for self collision checking

when changing numerical limits press enter to apply.

### DIFF
--- a/nixfiles/modules/home/macros/launch/arch/post-landing.nix
+++ b/nixfiles/modules/home/macros/launch/arch/post-landing.nix
@@ -12,7 +12,7 @@ let
   one = {
     pre = pre-shell {payload-name=task-name + " one"; need-rover=true;};
     terminals = [
-      {name = "Rover:Arm Control"; platform=rover; cmd="./ros2 launch arm_bringup control.launch.py local:=True | grep -v not.defined.in";}
+      {name = "Rover:Arm Control"; platform=rover; cmd="./ros2 launch arm_bringup control.launch.py local:=True";}
       {name = "Rover:Arm Can Sleuth"; platform=rover; cmd="./can_sleuth -o tui taipan";}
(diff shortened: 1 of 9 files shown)


Redecorating:
I think the wrist pitch feedback should be correct now. This means the wrist pitch limits should work although they are still not enabled for j5.

i also have made teleop arm log inputs by default

jacksonmacc (commented):
i havent run it yet but it looks good!

PR #969, opened 2026-03-14
feat: nomnomnomnomnom

vic-bart:
## Description

[nix-output-monitor (or nom)](https://github.com/maralorn/nix-output-monitor) is an incredibly useful visualiser for nix-* processes. Say goodbye to not knowing how long a job is taking to build, or how many jobs are being downloaded, built, or have finished, or not being able to tell which job corresponds to which line in the terminal output! We've been using `|& nom` for a while now, this PR just formalises that process by integrating nom directly into our aliases so nothing extra is required. 

## Changelogs

- Replaced every alias with `nix-build` or `nix-shell` with `nom-build` or `nom-shell` respectively.

## Screenshots

image

## Steps to Test

- `ws-build` (or any nix-build alias)
- `predict-shell` (or any nix-shell alias)

## Memes

image

Tags for Notion Integration:
tag:qol

### DIFF
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -49,8 +49,8 @@ in
           off = "sudo poweroff";
 
           # Nix CLI shortcuts
-          nova-build = "nix-build ${cfg.nixfileDir}";
-          nova-shell = "nix-shell ${cfg.nixfileDir}";
+          nova-build = "nom-build ${cfg.nixfileDir}";
+          nova-shell = "nom-shell ${cfg.nixfileDir}";
           ws-build = "${nova-build} -A pkgs.ros.nova-workspace";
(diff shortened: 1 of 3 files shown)


vic-bart:
`ws-build` works, `predict-shell` builds but running it gives:

image

But I guess this does not have to do with nom vs nix and more to do with me missing some part of the setup needed for ilmenite_ml? If so, this PR should be good to merge.

Redecorating (approved):
nice

vic-bart:
Running into an issue with nix-shell in unity-sim.
This is what nix-shell looks like:

image

This is what nom-shell looks like:

image

If I enter exit, it seems to fallback to running nix-shell:
image

I can enter exit again, which prints out what it should only print out when you create the shell:

image

PR #972, opened 2026-03-14
[low priority] blcmd_status_monitor for taipan

Redecorating:
it is set up so it publishes to a /arm/blcmds/xxxxx for arm and the /blcmds/xx for drive (same as previous). the gui only uses the drive one, it didn't look trivial to make a second copy of the widget in the navbar to do arm so I haven't looked into it much.

add it to control.launch.py in arm_bringup when not using mock hardware also move the can launch stuff to only be when not using mock hardware


Redecorating:
probably not gonna do more on this for the time being. i kinda want this to be part of blcmd hardware instead but that's a larger change so not for doing now

PR #973, opened 2026-03-14
Cameras2 custom pipelines

rotaryViper:
## Description

Added ability to change:
max-bitrate
video-caps

This lets us use h264. Not optimized, but fine for our use case. Better than none at all.

Should be a very stable change that shouldn't break anything major.

Tested working with 1 camera on the orin.

## Screenshots

site1-brandon-cam-1773496507118


V01DBREAKER (changes requested):
I would remove the configs youve created for testing, otherwise lgtm

V01DBREAKER:
Now that its after comp i reckon we dont merge this, keep cameras2 as is and just focus on cameras3. That way we have a known working camera stack.

PR #975, opened 2026-03-15
feat: Auto Cameras Page

FelicityMatt:
## Description

Auto only uses the mast cams so adjusting the view to only use three cols instead of four.

The AutoCameraPage is a direct copy of the normal CameraPage just with three cols, I'd like to do a major overhaul of the camera page anyway so not worth adding settings for it, especially so close to comp.

## Changelogs

- Created a dupe auto camera page

## Screenshots

image

## Memes

img_7242_720


TJ-Lee-M (approved):
👍

V01DBREAKER (approved):
LGTM!

PR #977, opened 2026-03-16
Configure ptpd service

tytia:
## Description

Configured to restart on failure indefinitely, added priority option to be able to preference base as master


Nova-Jetson:
LGTM if you have tested - Orlando

PR #979, opened 2026-03-18
feat: Update graph reading calculations

FelicityMatt:
## Description

Fixing a bug where previously only the first y value was used because they all had the same label.

## Screenshots

image


src/ros/nova-gui/nova-gui/src/components/science/NIRProbe/NIRProbeCalibration/NIRCalibrationCurveWidget.tsx
```diff
@@ -43,17 +43,26 @@ const NIRCalibrationCurveWidget: React.FC<NIRCalibrationCurveWidgetProps> = () =
...
+    // Compare based on index:
+    // match readings based on index
+    const indexPairedXY = readings[NIRProbeReadingType.PD1]
+      .filter((v, i) => i < readings[NIRProbeReadingType.PD2].length)     // cut pd1 readings to pd2 length
+      .map((v, i) => [v.data, readings[NIRProbeReadingType.PD2][i].data]) // list of [x, y] values
+    const zValuesScatter = indexPairedXY
+      .map(v => calibrationFunc(v[0], v[1]))
 
-    return {x: valuesScatter.map(v => v[0]), y: valuesScatter.map(v => v[1]), z: zValuesScatter, text: labels} as ScatterPlotData
+    // Compare based on labels:
```
FelicityMatt, line 54: Leaving this here in case we want to use it in the future -- the nir gui needs a bit of a refactor with the new design.

binudakal (approved):
lgtm

PR #980, opened 2026-03-18
Fix/new arm limits

Redecorating:
## Description

I have temporarily undone the changes to blcmd hardware so we can merge in the limit changes now and fix the rest later.


TJ-Lee-M (approved):
lgtm! happy to merge if these are the most recent on-arm tested params 🫡

PR #982, opened 2026-03-18
aliases, launch: enable ssh compression

### LINKED ISSUE #948: optimise ssh
### **Context**

we can turn on compression! i found out about this here https://infosec.exchange/@rebane2001/116200045748516097

there are probably other things we can do too.

### **TODO**

-

Redecorating:
the -C flag enables compression for ssh. very good for our use case as we want to minimise radio bandwidth usage


V01DBREAKER:
Thanks for doing it to launch scripts too :D lgtm

PR #986, opened 2026-03-19
Feat/custom remove in collision goals

hhoverall:
## Description

Add custom RemoveNearbyInCollisionGoals action node that mimics RemoveInCollisionGoals, but only for points within a certain distance of the rover. The point of this is to avoid prematurely removing goals that may be far away from the rover, resulting in the rover never reaching certain parts of the arena. With this new node, the rover should only abort the goal when it is sufficiently close, having traveled most of the way there and thus mapped out more of the arena.

## Changelogs

- Added RemoveNearbyInCollisionGoals.cpp and .hpp files
- Updated CMakeLists.txt to include this new node
- Created new BT nav_through_poses_remove_goals.xml that uses RemoveNearbyInCollisionGoals and GetCurrentPose 
- Updated bt_navigator.yaml to use this new BT by default. Previous BT (nav_through_poses) kept as a reference

## Steps to Test

1. run `mros2 launch auto_bringup everything.launch.py`. By default, bt_navigator.yaml is already configured to use the new nav_through_poses_remove_goals behavior tree. 

2. Go into NavThroughPoses mode and set some waypoints up. One of these should be inside an obstacle (about 5 grid spaces directly in front of the rover there is an obstacle you can use).

3. ???

4. Profit

## Memes

IMG_7453


tytia (commented):
Can you rename the BT to `nav_through_poses_remove_nearby_collision_goals`? I know it's a bit long but the regular `nav_through_poses` also removes goals so the current name is a bit ambiguous.

Also could you put the logic for getting the current pose inside the node instead of taking it in an input port? I think this would more closely match the style of RemovePassedGoals, plus currently there aren't any other nodes that would benefit from having a pre-calculated current pose on the blackboard. You can just copy how RemovePassedGoals does it on the jazzy branch: https://github.com/ros-navigation/navigation2/blob/jazzy/nav2_behavior_tree/plugins/action/remove_passed_goals_action.cpp

hhoverall:
I've been doing some more testing and currently, any goal inside a non-zero cost layer will be removed. This includes the inflated buffer layer around obstacles. To fix this I will need to add a param for determining the min cost threshold to consider a goal for removal. I believe this should just be another input port with a numerical parameter for min cost for removal.

tytia:
merged already a while ago from somewhere else

PR #987, opened 2026-03-19
Feat/prp

### LINKED ISSUE #627: automatic radio fallover/redundancy
- [x] try vxlan again with swapped slave order
- [x] use bridge so the setup works even if only one eth iface is present
- [x] make rules to rename first usb eth so we don't have to worry about which port it is in
- [ ] make ros dds profile and set with env var to make ros only do 10.0.0.0/23
- [ ] ~~figure out how to make network manager manage prp0 on base laptops so people can still edit their network settings~~
- [ ] ~~make netowrk manager default profiles for base, base shared?~~
- [ ] write example for using nixos specialisations to enable this networking configuration at boot but not by default (and write down how to remove it at runtime)

Tags for Notion Integration:
tag:software;tag:electronics

Redecorating:
## Description

support for parallel redundancy protocol so we can use both 900mhz and 5ghz radios at once, while having ssh and ros traffic not interrupted if one radio goes down.

PRP sends all packets down two seperate copies of the same network. for us this means that the packets are sent on both 900mhz radio and 5ghz radio. if a packet is lost over the air in one radio that's ok as the other one will come through. this means it should reduce packet loss a bit. The main advantage is that this means there is no downtime or need to reconnect if one radio goes out.

Currently the config gives you an ip in 10.0.0.X (redundant, default), 10.5.0.X  (only over 5ghz radio) and 10.9.0.X (only over 900mhz radio). The intended usage is that ros and ssh etc will go over 10.0.0.x, while in the gui you'd put rover ip as 10.5.0.X so cameras are streamed only over 5ghz (and you could have the other laptop or an incognito window stream maybe just one camera with 10.9.0.X so it goes over 900mhz).

the base laptops and orin would have an additional usb etherent device. there is a rule added to rename the first usb etherent to usbeth0.

closes #627 

## Steps to Test
run test vms:
`nova-build -A tests.networking`

add to your computer

```
nova.networking = {
  ethernetInterface = "XXX"; # on orin this is already defined.
  secondaryEthernetInterface = "usbeth0";
};
specialisation = {
  prp.configuration = {
    nova.networking = {
      prp = {
        enable = true;
        address = "0.XXX"; # or 1.xxx - must be unique to your computer. your ips will be 10.0.0.x, 10.5.0.x, 10.9.0.x
      };
    };
  };
};
```
then do `nixos-rebuild switch` and reboot, there should be a prp option in your bootloader menu. alternatively you should be able to do `sudo /run/current-system/specialisation/prp/activate` maybe?

TODO: configure 900mhz radios


banksia-software:
## BASE ##
AT+SSID=nova900mhz
AT+KEYMGMT=WPA-PSK
AT+PSK=cafeb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0b
AT+TXPOWER=20
AT+MODE=ap
AT+BSS_BW=8
AT+FREQ_RANGE=9200,9280

## ROVER ##

AT+SSID=nova900mhz
AT+KEYMGMT=WPA-PSK
AT+PSK=cafeb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0bb0b
AT+TXPOWER=20
AT+MODE=sta
AT+BSS_BW=8
AT+FREQ_RANGE=9200,9280



view all config

AT+WNBCFG

station/client

AT+PAIR=0/1

if distance is greater than 3km increase ack timeout!


https://portal.powertec.com.au/industry-resources/countries-territories/asia-pacific/australia/spectrum/au-915-928-mhz

1W max tx power. this is 30dBm

banksia-software:
don't merge yet, shm transport within the same computer doesn't work? - orlando

Redecorating:
can't reproduce with vm tests, maybe i hadn't rebooted when i was trying it.

Redecorating:
did not impact ros for running arm on rover (aside from us forgetting to change the ip in the gui for cameras). It handles the case where the usb eth isn't plugged in.

need to add the mjpeg_bpp quirk to nixfiles (i swear i did this :/). I'll make a pr for that in a minute

rover 900mhz radio is dead? :( i thought it powered on fine with ac adapter :/ double check this future me.

PR #988, opened 2026-03-20
feat: Add cut auto mount

V01DBREAKER:
## Its not about the size (211mm reduction!), its about how you use it

image

I tested this with urdf viewer in vscode but plz check in rviz first and make sure transforms look correct

Tags for Notion Integration:
tag:autonomous;tag:rviz;tag:ready_for_review


tytia:
hell yeah looks great Anthony, just gonna make some further urdf modifications before I merge this (arg for using old vs new clam, removing oaks)

PR #990, opened 2026-03-20
we rovers have no limits

Redecorating:
make the trigger that was for entering task space just enter no limits mode

## Steps to Test

you can test this with mock.launch.py


Redecorating:
Let's Charge Together With Full Battery

jacksonmacc (approved):
looks good to me!

Redecorating:
ik uses that button now sooo

PR #991, opened 2026-03-20
fix: plotly bundle

binudakal:
## Description

importing the entire plotly bundle causes oom errors on some devices when building the gui, so importing the minimal 2d and 3d bundles separately should prevent this


FelicityMatt:
lgtm if you've tested

PR #1000, opened 2026-03-24
remove prp vlans and dds config

Redecorating:
shm dds didn't work? might only be when ethernet isn't plugged in? for now just remove vlans so we don't need to reconfigure dds


Redecorating:
one thousand!

PR #1001, opened 2026-03-25
Feat/remove nearby in collision goals

hhoverall:
## Description

Node takes input param of distance and cost threshold. CURRENTLY any point inside an obstacle as well as the inflated (cyan) layer around it has a cost value of 99. The darker blue layer the interpolates between 99 and 0 the further away it is from the cyan.

By default, this is set up with a threshold of 99 (exclusive). So will not remove any points below this value. You can lower this to preferentially delete nodes that are "too close" to the cyan inflated layer, eg by setting it to 80 or something.


banksia-software (approved):
lgtm, merging - @tytia

PR #1002, opened 2026-03-25
feat: auto blcmd reset

Glass2721:
## Description

automatically reset blcmds (currently enabled only for the auto task in arch) when an error is received. parameters are highly flexible, so feel free to ask Jonathan Jia how any of this works.

change params for this feature in `nova.yaml` and `auto.yaml` in drive_bringup (in the `blcmd_status_monitor` section of the file)


banksia-software (approved):
lgtm, merging before comp - @tytia

Issue #1005, opened 2026-03-29
Create MultiSensorHardware


YahyaMuayyiduddin:
### **Context**

Create a multisensorhardware for hardware that needs to send to multiple state interfaces
    * reads from one can message → multiple state interfaces
    * has a list of state interface names/units/types (e.g. temp1, temp2)
    * has a list of corresponding can message conversions.

### **TODO**

-

Tags for Notion Integration:
tag:software;tag:science

PR #1006, opened 2026-03-29
feat/MultiSensorHardware

YahyaMuayyiduddin:
## Description

MultiSensorHardware interface to read from 1 can message and send data to multiple state interfaces.
Has a list of state interface names, their units, and their corresponding data conversion functions.
For example, if a can device sends a payload with readings from multiple sensors, [sensor1 sensor2 sensor3 sensor 4],
each corresponding data conversion function should extract their specific data, and convert it, to be put onto its corresponding state interface.

Example:

.with_hardware("bme_sensor", MultiSensorHardware,
    can_id=0x123,
    hardware_names=["temp1", "temp2", "pressure"],
    hardware_units=["temperature", "temperature", "pressure"],
    interpret_data_list=[
        lambda data: int.from_bytes(data[0:2]) * 0.01,  # bytes 0-1 → temp1
        lambda data: int.from_bytes(data[2:4]) * 0.01,  # bytes 2-3 → temp2
        lambda data: int.from_bytes(data[4:6]) * 0.1,   # bytes 4-5 → pressure
    ]
)
#1005 

## Changelogs

- AddedMultiSensorHardware

## Memes

image

### DIFF
--- a/src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/MultiSensorHardware.py
+++ b/src/ros/rover/nova_generic/python_control2/python_control2/hardware_interfaces/MultiSensorHardware.py
@@ -24,14 +24,12 @@ class MultiSensorHardware(HardwareInterface):
 
     def __init__(self, contexts: Contexts,
                  can_id: int = 0,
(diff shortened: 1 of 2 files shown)


FelicityMatt (commented):
lgtm

PR #1024, opened 2026-04-08
gui: nix: rebuild node modules less

Redecorating:
by default, mkYarnPackage sets its packageJSON and yarnLock to src + filename. this means that when those arguments are passed down to the builder for the node modules, its in the form of /nix/store/{entire gui source code}/package.json. that means any change to gui code changes the nix store path for the node modules builder so it has to rebuild it even though the particular files (package.json + yarn.lock) haven't changed.

work around this by copying those files to the nix store seperately. this means that their nix-store hash is only based on the individual files' contents not the whole gui source code, and you rebuild node-modules only when you actually changed the modules.

https://github.com/NixOS/nixpkgs/blob/5f7ca9ab52cf303ec12876e19150cb079046fa0b/pkgs/development/tools/yarn2nix-moretea/default.nix#L312-L313

### DIFF
--- a/src/ros/nova-gui/nix/packages/gui/default.nix
+++ b/src/ros/nova-gui/nix/packages/gui/default.nix
@@ -33,6 +33,11 @@ in
 mkYarnPackage {
   name = "gui";
 
+  # Make sure that the node modules derevation doesn't have the whole source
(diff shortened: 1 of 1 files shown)


Redecorating:
#2^10

V01DBREAKER (approved):
LGTM I also hate node modules

PR #1027, opened 2026-04-10
fix: Gui Linting Errors Begone!

V01DBREAKER:
## Description
I fixed or ignored all the linting issues. Enjoy ;)

## Changelogs
Rewrites cartographer map tiler to avoid one of them which also improves its functionality when switching between maps.
WebGL's entire structure abuses ref passing which without a full rewrite would be hard to fix so i've ignored most of it
Any useEffect which had a setState function in it created an error and most of these were moved into useEffectEvent which is essentially the same thing but won't cause rerenders. See [here](https://react.dev/reference/react/useEffectEvent) for more details.

## Steps to Test
I was able to test the camera page changes, cartographer changes and simple gui components but science is a black box so please when reviewing focus your effort on science!

Run gui in standard way e.g `gui-shell` -> `gui-link` -> `gui-run` | `gui-rosbridge` | `tileserver` | ros nodes that send data

## Memes

image
This PR better not be bikeshedded istg

Tags for Notion Integration:
tag:nova-gui


src/ros/nova-gui/nova-gui/src/components/drive/DriveControlWidget/DriveControlWidget.tsx
```diff
@@ -34,7 +34,7 @@ const DriveControlWidget: React.FC<IDriveControlWidgetProps> = (
     "3": "Excessive BLCMD Errors",
     "4": "Gamepad Disconnected",
   };
-  const suppressLockedReasonToast = [1, 2];
+  const suppressLockedReasonToast = useMemo(()=>[1, 2], []);
```
Glass2721: what why did it want this

src/ros/nova-gui/nova-gui/src/components/drive/DriveControlWidget/DriveControlWidget.tsx
```diff
@@ -44,7 +44,7 @@ const DriveControlWidget: React.FC<IDriveControlWidgetProps> = (
     if (!suppressLockedReasonToast.includes(driveInfo.locked_reason) && driveInfo.locked) {
       toast.error(`Drive Locked due to: ${currentLockedReason}`);
     }
-  }, [driveInfo.locked_reason, driveInfo.locked]);
+  }, [driveInfo.locked_reason, driveInfo.locked, currentLockedReason, suppressLockedReasonToast]);
```
Glass2721: ig this is okay, but janky (the others don't need to trigger this useeffect)

src/ros/nova-gui/nova-gui/src/components/maps/Cartographer/components/AutoArrivedPopup.tsx
```diff
@@ -22,11 +22,13 @@ export const AutoArrivedPopup : React.FC<AutoArrivedPopupProps> = () => {
     bifrost.syncWithTopic()
   }, [bifrost]);
 
+  if (autoStatus === IRosNovaInterfacesStatusConst.ARRIVED_SUCCESSFULLY && lastStatus != autoStatus) {
```
FelicityMatt, line 25: I think I want this in a useEffect, the whole point is that it only pops up one time when the auto status changes it changes, I don't want it to be up always when the status is successful

FelicityMatt: ok nvm it seems to still work.

FelicityMatt (changes requested):
Biggest issues are: 
- the 360 cam has lost the example image and doesn't seem to work, I just get a blue screen, 
- the cartographer won't allow me to place down points

happy to just ignore those issues, not worth spending our time on things that work fine imo, but up to you if you want to spend the time.

science all appeared to work!

src/ros/nova-gui/nova-gui/src/components/drive/DriveControlWidget/DriveControlWidget.tsx
V01DBREAKER: The 'suppressLockedReasonToast' array makes the dependencies of useEffect Hook (at line 47) change on every render. Move it inside the useEffect callback. Alternatively, wrap the initialization of 'suppressLockedReasonToast' in its own useMemo() Hook.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)

src/ros/nova-gui/nova-gui/src/components/drive/DriveControlWidget/DriveControlWidget.tsx
V01DBREAKER: React Hook useEffect has missing dependencies: 'currentLockedReason' and 'suppressLockedReasonToast'. Either include them or remove the dependency array.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)

V01DBREAKER: After some more research I have learnt that a useEffect is actually not the best way to do this. Since useSelector pulls information from the state and forces a render on the state changing you dont even need the useEffect

however it does mean u have to have multiple useSelectors for property of the redux state:
image

V01DBREAKER: I tested this with the ros topic so its working as intended!

src/ros/nova-gui/nova-gui/src/components/drive/DriveControlWidget/DriveControlWidget.tsx
V01DBREAKER: this has been removed with removal of useEffect

src/ros/nova-gui/nova-gui/src/components/drive/DriveControlWidget/DriveControlWidget.tsx
Glass2721: so I've placed useEffect here to only trigger the check/notification on a re-render if one of the states (and no other state) related to drive locked reason has changed (i.e. it's been updated by a new message from teleop drive). Doing it without useEffect, I fear, risks the notification being triggered on any re-render, not just a re-render with updated relevant state (i.e. locked reason and locked state) from teleop drive (implying a new message was received). Consider the following situation:
1. drive is currently locked with reason excessive blcmd errors
2. controller is disconnected suddenly
3. drive control widget is re-rendered, triggering a notification that drive has been locked due to blcmd errors (which is undesirable, because this notification would have already been shown, and would be wrong because the locked state has not changed from locked -> unlocked, or locked reason has not changed to a locked reason that is not suppressed)
wdyt? (idk why the linter is complaining about not having all state as dependencies, I suspect it may be because the pattern I'm using here is sus, but this may be a consequence of how bifrost handles state; idk)

sorry for writing such a long reply, if i had more time i would've written a shorter comment

src/ros/nova-gui/nova-gui/src/components/drive/DriveControlWidget/DriveControlWidget.tsx
Glass2721: If we're keeping useEffect, then "Move it inside the useEffect callback" is probably best bet (don't need to throw useMemo at this trivial thing). IMO this is an undetected bug on my part for not handling useEffect dependencies properly for mutable objects.

src/ros/nova-gui/nova-gui/src/components/maps/Cartographer/components/AutoArrivedPopup.tsx
V01DBREAKER: yes the useEffect is unnecessary as it also causes a rerender which is unneeded as the correct conditionals on the first render pass will also solve the issue!

V01DBREAKER:
Hey @FelicityMatt i've rewritten the cartographer and 360 cam parts and they should work now!

Also something i've noticed is that we use useEffect for transforming data we recieve from useSelector which while works in not optimal and hence all the linting errors. useSelector actually triggers the renderer so you can actually just have good conditionals to ensure correct data transformation and renders. to get a bit more complex you can also modify the equality check that useSelector uses! But i've mainly replaced it with multiple useSelectors which helps prevent render loops.

Turns out useEffect rerenders the entire page's content when activated! This is very inefficient just to convert a number to a string etc.

Checkout this article for more
https://dev.to/imamdev_/hello-react-goodbye-useeffect-i-hope-4mf1

src/ros/nova-gui/nova-gui/src/components/drive/DriveControlWidget/DriveControlWidget.tsx
V01DBREAKER: In this case we should add an additional condition to check if the controller is indeed connected or not and whether or not to send the toast accordingly. useEffect should not be used for this as will trigger a rerender when used it just won't trigger the toast cause its been hidden in the useEffect hook. Thank you for the long reply with your thinking i appreciate it :D I think your logic just needs to be shored up, no need to use this hook and create excessive load on the GUI

V01DBREAKER: Also the change i made to the driveInfo useSelector part actually mitigates this provided the correct conditionals!

V01DBREAKER:
More people should review this, i like this discussion and learning a lot about how react is designed:
image

src/ros/nova-gui/nova-gui/src/components/drive/DriveControlWidget/DriveControlWidget.tsx
Glass2721: You've managed to nerd snipe me with react stuff, will take a look later when I don't have more pressing nova work. If you want to resolve this quickly, I happy for @FelicityMatt to make the final call on these changes

FelicityMatt: I'm happy to leave it as how @V01DBREAKER has updated it, I had the same scenario with an auto cartographer thing where I had a useEffect that I only wanted to trigger when certain things changed but the if statement appeared to work for it (I tested it a bunch), I think let's try this approach and if it doesn't work we can revert or fix but seems to be better so :shrug:

FelicityMatt (changes requested):
The two things I brought up are looking good! found another issue tho on the 360 cam page.

image

This is supposed to calculate the height based on distance and angles, it should work by filling in one angle with shift click and the other with ctrl click or entering in the numbers and running the calculations, neither is working.

bonus if you want to make it look a little nice / add some comments about shift and ctrl click 👀

V01DBREAKER:
> The two things I brought up are looking good! found another issue tho on the 360 cam page.
> image
> 
> This is supposed to calculate the height based on distance and angles, it should work by filling in one angle with shift click and the other with ctrl click or entering in the numbers and running the calculations, neither is working.
> 
> bonus if you want to make it look a little nice / add some comments about shift and ctrl click 👀

should work as intended and i made it look a bit nicer + instructions

FelicityMatt (approved):
LGTM

Also I updated the calc layout since imo I wanted to minimise it's height since there was so much width to work with, hope you don't mind

Image

PR #1032, opened 2026-04-14
feat: Science Pumps

### LINKED ISSUE #1021: Update Pump control
### **Context**

Science wants to update the pumps, moving from a timed continous motor to a positional servo I think?

This will need some discovery work and confirmation with science about their intentions.

Some pumps will be QCMD - should get them working, can use the existing code - will need to update the gui with more options

Stepper implementation is lower priority for now

### **TODO**

- Update the control code (migrate to PC2)
- Update the gui which their desired control.

Tags for Notion Integration:
tag:software;tag:science

FelicityMatt:
## Description

Uplift the science pumps code to accommodate more pumps

addresses: #1021 

> [!NOTE]
> this PR just address the updated QCMD pumps, science wants to add stepper pumps but I will address that separately once they are more sure about it. Also this does not have finalised timings or can ids, I will update those separately. 

## Changelogs

- Update the Pumps widget to make it shorter, also moved from segmented picker to drop down since there are now too many options.
- Migrated the ros2 node to pc2
- Moved the pumps ros network stuff to services and topics instead of an action since it is more reliable and fits better with the pc2 update loop (otherwise it blocks the loop to give continous feedback and I didn't wanna deal with that)
- Added a QCMD option to not spam 0s, default is off.

## Screenshots

image

https://github.com/user-attachments/assets/21cddc64-5dd1-45fd-a255-dd58a6f15fd8

## Steps to Test

run the ros2 node `ros2 run science pumps.py`

Note: will need to run `gui-link`

## Memes

Behold a cache:
image


src/ros/rover/nova_generic/python_control2/GUIDE.md
```diff
@@ -0,0 +1,755 @@
+# How to Write a Python Control2 System
```
FelicityMatt, line 1: Got claude to generate this, thought it might be helpful for newcomers and other agents to read to gain context.

V01DBREAKER:
Sorry @FelicityMatt i've undone the rebase and will let you handle it

PR #1033, opened 2026-04-15
nova arm controller: in position mode limit all joints together

Redecorating:
tested this on the arm today and wrist pitch/yaw didn't seem to do the weird stuff anymore. j3 had bad resolver so i couldn't fully test it.


Redecorating:
this will make the arm slow down all joints when it is near the singularity, so it should handle it better.

PR #1045, opened 2026-04-18
feat: YOLO in GUI

### LINKED ISSUE #1013: YOLO in GUI
### **Context**

From @tytia on Slack
CRITICAL FEATURE FOR AUTO:
YOLO detections in base (GUI)

We want to run YOLOv8 (or another version?) in base on the camera feeds in the GUI and show the bounding boxes of detections to relay to the judges. The rover should stop once a there is a reliable detection. This can be done cancelling auto's navigation through ROS 2, more specific details to come later (should be really simple, I just forget whether it's an action or a service call that needs to be sent).

We would like for the detections to be filtered to make sure they're reliable (as there can be false detections). Last year this was done on auto's side in the behaviour tree, but I'd like for it to be done on the base/GUI side this year to make it simpler for auto. Last year we took in all raw detections and determined whether a detection was valid based on the proximity and time of detections. The criteria was "n detections in the last m seconds within a cluster", with a cluster being defined with a radius. I suppose with 2D detections this radius will now need to be a radius of pixels? The default we set last year was n = 2, m = 5 and cluster radius = 1.5 (m). We don't have to do filtering the same way this year, but [here](https://github.com/MonashNovaRover/nova/blob/master/src/ros/rover/auto/nova_behavior_tree/include/nova_behavior_tree/condition/detected_condition_base.hpp) is our code for it last year.

We (auto) will try to help you guys out on this as much as possible, but we'll need to do a bit of learning as most of us are not familiar with the GUI

### **TODO**

-

Tags for Notion Integration:
tag:autonomous;tag:software;tag:nova-gui

FelicityMatt:
## Description

CRITICAL FEATURE FOR AUTO:
YOLO detections in base (GUI)

We want to run YOLOv8 (or another version?) in base on the camera feeds in the GUI and show the bounding boxes of detections to relay to the judges. The rover should stop once a there is a reliable detection. This can be done cancelling auto's navigation through ROS 2, more specific details to come later (should be really simple, I just forget whether it's an action or a service call that needs to be sent).

## Changelogs

- Added YoloDetection code that can be used within a `YoloCameraComponent`.

## Screenshots

image

## Memes

image


tytia (approved):
Highkey idk what I'm looking at but if it breaks something so be it we can rollback 🤷 

I just wanna merge stuff into master ASAP so we don't have a repeat of the auto merge nuke of ARCh 2026

PR #1050, opened 2026-04-19
can_sleuth: refuse to emulate when not on vcan

### LINKED ISSUE #1047: can_sleuth: don't emulate if real device is present
### **Context**

apparently if can_sleuth emulates when there's actual hardware it kills the blcmds and they need reset? maybe it tries to send telemetry with the same canid as what the blcmd sends at the exact same time causing the can controller to go into error mode?

### **TODO**

- option 1: make it check `readlink /sys/class/net/{canX} | grep virtual` or smth similar with python not bash to ensure it is running on vcan when it does the emulator (probably do this in systems.py
- option 2: make the emulators listen to can messages from the actual device for a second before sending telemetry and if it hears from the actual device then die. (probably harder to do)

Redecorating:
emulating when there's a real blcmd present makes the blcmd require a reset

closes #1047


Redecorating:
image
tested on agx

PR #1051, opened 2026-04-19
nixfiles: allow nova to use rt fifo scheduler

### LINKED ISSUE #1043: ros2control use rt scheduling
### **Context**

[ros2_control_node-6] [WARN] [1776428397.380179396] [controller_manager]: Could not enable FIFO RT scheduling policy: with error number <1>(Operation not permitted). See [https://control.ros.org/master/doc/ros2_control/controller_manager/doc/userdoc.html] for details on how to enable realtime scheduling.

if we improve this then maybe it will run behind less - sometimes when there's high cpu load the controller manager falls behind. 

proper RT needs a different kernel but we can still get a better scheduler for ros2control without that from what i have read in the link above.

### **TODO**

- 
https://control.ros.org/master/doc/ros2_control/controller_manager/doc/userdoc.html

https://search.nixos.org/options?channel=25.11&include_modular_service_options=1&include_nixos_options=1&query=loginLimits&show=option:security.pam.loginLimits

also check that gstreamer doesn't also use this cause if we end up with both ros2control and cameras having high priority then this may not help (although maybe you'd be able to set different priorities within the FIFO scheduler)

Redecorating:
ros2control complains about not being able to do this. this should help with the responsiveness of ros2control.

to confirm it works:

chrt -a -p $(pgrep ros2_control_no) | grep -A1 FIFO

closes #1043

### DIFF
--- a/nixfiles/modules/nixos/common/users/default.nix
+++ b/nixfiles/modules/nixos/common/users/default.nix
@@ -40,12 +40,36 @@
             wheel.name
             video.name
             dialout.name
+            realtime.name
           ];
         };
       }
     ];
 
     nix.settings.trusted-users = lib.optional config.nova.users.nova.enable config.users.users.nova.name;
+
+    # Allow RT FIFO Scheduler, locking memory not in swap for ros2control
+    users.groups.realtime = {};
+    security.pam.loginLimits = [
+      {
+        domain = "@realtime";
+        item = "rtprio";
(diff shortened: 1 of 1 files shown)


Redecorating:
image
seems fine on agx

banksia-software:
https://github.com/MonashNovaRover/nova/issues/1043#issuecomment-4287680647

PR #1053, opened 2026-04-20
No more bunny hops :rabbit: (+ other drive fixes inc. no more teleop lag!) 

tytia:
## Description

Fix stopping instantly when connection is lost and be more lenient towards transient connection dropouts.

To be merged when tested IRL!

Edit:
Okay I ended up applying some other improvements/fixes on this branch so there's a couple more changes than originally anticipated :sweat_smile: 

Additional changes are listed in the changelogs section :arrow_down: 

## Changelogs

- Don't bypass derived controller limiters when sending zero commands after losing connection
- Stricter command timeout at 0.1 (can be adjusted after further testing at long range with unstable connections)
- Toned down speed limits in `nova.yaml` and `auto.yaml` after the 2.4 multiplier removal in the firmware
- Changed min_interval in `ros2_control.xacro` to 233
- Improved logging in drive controllers; it's now coloured and tells you when connection is lost (both temporarily and for extended periods of time)
- Changed QoS profile for drive controllers and teleop
  - Best effort with depth 1 for drive controllers makes sense and also seems to be fairly standard (it's what ackermann_steering_controller uses, which is not written by us)
  - Reliable with depth 1 for teleop, as it seems teleop generally uses reliable?
- **IMPORTANT** Decouple teleop_drive_joy's publishing from the publish rate of ros2 joy, this resolves the intermittent lag issue. Currently set the publish rate to 20 Hz
- Keep publishing zero commands when controller is locked, teleop should always be publishing unless the controller is physically disconnected

## Screenshots

image

## Steps to Test

```
# start vcan
can start vcan0

# launch drive
ros2 launch drive_bringup drive.launch.py

# launch teleop
ros2 launch teleop_drive_joy teleop.launch.py
```

### DIFF
--- a/src/ros/rover/drive/drive_controllers/nova_drive_controller_base/src/nova_drive_controller_base.yaml
+++ b/src/ros/rover/drive/drive_controllers/nova_drive_controller_base/src/nova_drive_controller_base.yaml
@@ -109,9 +109,11 @@ nova_drive_controller_base:
     default_value: 999.0 # Seconds
     description: "Timeout to receive cmd_vel commands from base."
 
+  # Set to a higher value to be more lenient towards transient connection dropouts,
+  # however this also means the rover will take longer to stop when connection is truly lost.
   cmd_vel_command_timeout:
     type: double
-    default_value: 0.5 # Seconds
+    default_value: 1.0 # Seconds; teleop publishes at 50Hz i.e. every 0.02 seconds
     description: "Timeout since last commanded velocity from cmd_vel."
 
   publish_commanded_velocities:
--- a/src/ros/rover/drive/drive_controllers/nova_drive_controller_base/src/nova_drive_controller_base.yaml
+++ b/src/ros/rover/drive/drive_controllers/nova_drive_controller_base/src/nova_drive_controller_base.yaml
@@ -113,7 +113,7 @@ nova_drive_controller_base:
   # but the rover will take longer to stop when connection is truly lost.
   cmd_vel_command_timeout:
     type: double
-    default_value: 0.1 # Seconds; teleop publishes at 50Hz i.e. every 0.02 seconds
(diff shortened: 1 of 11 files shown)


src/ros/rover/drive/drive_controllers/nova_drive_controller_base/src/nova_drive_controller_base.yaml
```diff
@@ -109,9 +109,11 @@ nova_drive_controller_base:
     default_value: 999.0 # Seconds
     description: "Timeout to receive cmd_vel commands from base."
 
+  # Higher values are more lenient towards transient connection dropouts,
+  # but the rover will take longer to stop when connection is truly lost.
   cmd_vel_command_timeout:
     type: double
-    default_value: 0.5 # Seconds
+    default_value: 0.1 # Seconds; teleop publishes at 50Hz i.e. every 0.02 seconds
```
Glass2721: iirc it's 20hz minimum, and up to 1000hz as required (as that's the frequencies for which joy node publishes messages at)

src/ros/rover/rover_description/banksia/urdf/ros2_control.xacro
Glass2721, line 1: I assume the correct min interval for drive is 233 (as that's what the changes to xacro params does). maybe better to rename `gear_ratio` xacro param to `drive_gear_ratio` cause we don't use it for the pivots

src/ros/rover/drive/drive_controllers/nova_drive_controller_base/src/nova_drive_controller_base.yaml
Glass2721: idk if this new value is too tight; test at long range with bad radios to see performance

tytia:
drive telemetry still pending to be tested after the 122 -> 233 `min_interval` update

PR #1056, opened 2026-04-20
nixfiles: fastlivo2: don't make derevation depend on $HOME

Redecorating:
not ideal to hard code this to /home/nova, but rn hydra builds a different derivation because its not nova building it so everyone has to build fastlivo themselves till someone manually copy closures each version to hydra.

gonna see if hydra builds this right


Redecorating:
image

tytia (approved):
LGTM, thanks

PR #1057, opened 2026-04-20
arm: make fingers fixed in rviz

Redecorating:
## Description

need to test if this messes up the joysticks controlling the fingers and poker.

## Memes

Tags for Notion Integration:
tag:ready_for_review


TJ-Lee-M (approved):
yipeeee thankyou :)

PR #1063, opened 2026-04-21
fix: auto cartographer map goals modal

V01DBREAKER:
Linting Error Fix accidentally broke this feature so this PR is two fold.
The first commit changes auto homepage icon in side bar
Second commit undos the linting change which broke it
Third commit rewrites the modal so that redux state store and react state are separated which removes the linting error and keeps the functionality.

To test:
`gui-shell then `gui-link` then `gui-run` and navigate to localhost:5173/urc/autonomous-navigation
`tileserver`
`gui-rosbridge`

When goals are published you can observe the correct order in the console as it also logs to web console (F12 menu)
Selected points also reset when reopening modal
Points should update the modal as since points are part of redux state it will force a rerender through the library.

Tags for Notion Integration:
tag:nova-gui

### DIFF
--- a/src/ros/nova-gui/nova-gui/src/components/auto/Cartographer/components/AutoStatus.tsx
+++ b/src/ros/nova-gui/nova-gui/src/components/auto/Cartographer/components/AutoStatus.tsx
@@ -8,14 +8,12 @@ import {RosTopic} from "../../../../ros/topics/rosTopic.ts";
 interface AutoStatusProps {
 }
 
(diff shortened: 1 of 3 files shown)


FelicityMatt (commented):
I made a fixed for the colours for the auto status

the modal is also a little buggy is that intended behaviour?

https://github.com/user-attachments/assets/10a0908c-e70b-4789-8701-7bf0818de6d2

V01DBREAKER:
Fixed!

FelicityMatt (approved):
lgtm

PR #1064, opened 2026-04-22
feat: drone gps in cartographer

henrylaw893:
## Description

Adds drone to the Cartographer GUI and installs mission planner so that the drone's flight plan can be updated live during the task

The handheld drone controller will send telemetry data to the cartographer laptop over wifi. The cartographer laptop will then use mavproxy.py to open two more internal connections to the drone. One of these will be used for mission planner, and another of these connections will be used along with the pymavlink library to continually get the GPS position and publish it to the ros2 topic /gps_drone/fix_custom. The cartographer GUI then subscribes to this topic and uses it to update the latitude, longitude, altitude and heading of the drone in the GUI

## Screenshots

image

[Screencast From 2026-04-19 15-01-52.webm](https://github.com/user-attachments/assets/2c70d6c2-5dcf-4307-b95c-f1cca731933a)

## ROS

Adds node DroneGPS from drone_gps.py
Adds topic /gps_drone/fix_custom which is a GPSData message type, containing latitude, longitude, altitude and heading

## Steps to Test

1. ```ws-build -o ~/Builds/master```
2. Connect to rover via ethernet
3. Power on the drone and handset controller
4. Connect to the handset controller wifi - something like "ExpressLRS TX Backpack" (password is ```expresslrs```)
5. Configure network settings so that connecting to wifi does not kill rover connection. Do this by changing the laptops wifi ip to ```10.0.0.2/30```. Restart wifi (turn on and off again) if this doesnt fix it. If it still doesnt work, change the ethernet to ```10.0.0.X/23```, where X is a number that isnt being used by any other profiles
6. run ```~/Builds/master/launch/run-drone```
7. Mission planner will run automatically. To connect to the drone, select UDP, press connect, and select port ```14550```
8. Run GUI as normal

## Meme
meme


src/ros/rover/drone/drone_gps/params/drone_gps.yaml
```diff
@@ -0,0 +1,2 @@
+/drone_gps:
```
FelicityMatt: is this needed if it is empty?

FelicityMatt (commented):
Overall looking good!

Just the one comment and then moving everything to where the existing gps lives, having a whole drone_gps package is excessive imo.

FelicityMatt (approved):
Made two small updates: 
- removed unused gui import that was failing linting
- renamed the launch file to `gps_drone` to be consistent with the others.

The gui works as expected without testing with moving gps, happy to merge and I can test later today.

PR #1068, opened 2026-04-23
Cameras3 ros image sink pixel format selector

rotaryViper:
## Description

Selecting a format will pass it to cameras3 v4lrosimage.
Also defaults to vp8 rather than h264, to be more in line with cameras2.

## Memes

Tags for Notion Integration:
tag:cameras


binudakal (approved):
tested and works, lgtm

PR #1080, opened 2026-04-25
feat: Configure Hall Effect sensors

### LINKED ISSUE #1041: Configure Hall Effect Sensors
### **Context**

title

### **TODO**

-

Tags for Notion Integration:
tag:software;tag:science

FelicityMatt:
## Description

Configuring the hall effect sensors on the augers and carousel.

Also fixed the local param for the launch files and added a can bus override so that we can easily switch canbuses in one place

## Changelogs

- Updated the auger hall effect camera component
- Added an auger config that includes publishing the sensors
- Added a temp carousel telemetry to the carousel widget (I still want to make improvements but will leave to after we have mvp)
- Updated the carousel code to use the hall effect sensors.

## Screenshots

image

image

## Steps to Test

run the gui and urc launch file

## Memes

image

Tags for Notion Integration:
tag:science


V01DBREAKER (approved):
Builds and runs with expected behaviour in GUI 

LGTM

PR #1082, opened 2026-04-25
feat: litmus strip servo control

binudakal:
## Description

Adds control for dipping the litmus strips into the 'shot glasses', including twitching and timed dips with both twitch amount and duration customisable. 
Note that currently the input to change the twitch step in is in degrees, it may be possible to figure out what the linear distance equivalent is but i've just left it as is for now to avoid overcomplicating things

## Screenshots

image

## ROS

services:
`/science/litmus_dipper/dip`: start dipping
`/science/litmus_dipper/stop`: stop dipping
`/science/litmus_dipper/twitch`: twitch the platform by a certain amount

topics:
`/science/litmus_dipper/status`: whether or not the dip is running, elapsed dip time, and the target dip time

## Steps to Test

1. run gui
2. `mros2 launch science_bringup urc.launch.py`
3. `candump can1 | grep 0E0`


V01DBREAKER (approved):
adjusted servo angle. works in gui lgtm

PR #1084, opened 2026-04-25
feat: auto-urc

tytia:
## Description

Latest tested auto changes.

## Changelogs

- Updated aliases
  - `launch-everything` is now `launch-sim`
  - Added `launch-gazebo` for `~/Builds/active/bin/ros2 launch auto_bringup gazebo.launch.py`
  - Renamed `start_arch` and `start_urc` for `start_auto_arch` and `start_auto_urc`
- Fixed IMU covariance patch for `livox_ros_driver2`
- Updated URC yolo model
- Updated Auto Cartographer GUI
  - Goal types are sent per service call as opposed to per goal, but you can still specify each goal's type when placing and by default the service sends the type of the last goal
  - Can now specify search radius per goal
  - Hooked up to auto stack with updated `CartographerCommand` service, tested and works
- GUI now publishes detections onto the `/yolo/object_detections` topic using new custom `Detections2DArray` msg
- `ObjectDetectedCondition` BT node is now connected the the aforementioned topic, tested and works, though improvements can be made (see TODO in source file)
- Renamed `gazebo` launch arg to `sim` in all auto launch files
- Added heading as IMU in gazebo
- Updated `RemoveInCollisionGoals` to work with costmap updates not just full costmaps (TODO update `RemoveNearbyInCollisionGoals` and `SnapInCollisionGoals` to use the same method of costmap consumption)
- Don't send full local costmap (only updates)
- Don't publish global costmap
- Increased the update and publish frequency for local costmap to 20 and 5 hz respectively
- Increased lidar publish rate to 20 hz
- Retuned MPPI configs, fast = 1.0 m/s max and turbo = 1.25 m/s max
- Disabled Prefer Forward Critic in MPPI
- Renamed URCNavigator to URCThroughPoses
- Tuned EKF to incorporate GPS/magnetometer heading for yaw
- Updated `everything.rviz` config
- Distributed and completely restructured auto start scripts `start_auto_urc.py` and `start_auto_arch.py`, tested IRL and works
  - Now lives in its own package directory under auto, `auto_start`
- Fixed offset radius in `SetDetectionGoalAction`, declared `goal_offset_radius` in `bt_navigator.yaml`
- Updated drive limits (stricter acceleration limits and lower max angular velocity)

## Steps to Test

```
# after nixos-rebuild
launch-sim
```


tytia (commented):
lgtm 👍

PR #1085, opened 2026-04-25
Feat/urc cache

### LINKED ISSUE #1020: Configure Caches
### **Context**

Cache control is now continuous

### **TODO**

- remove widget from gui
- do pc2 continous system use joysticks
  - Continuous servo 
  - I think the best option would be using the joystick thumb axis, do one cache for each for now

Tags for Notion Integration:
tag:software;tag:science

YahyaMuayyiduddin:
## Description

Migrated caches to python control 2. The left and right caches are controlled using the left and right joystick thumb axis' respectively.

#1020 

## Changelogs

- Added urc_cache.py
- Removed cache gui widget
- Edited urc.yaml
- Edited teleop.yaml
- Edited joystick.config.yaml


V01DBREAKER (approved):
Fixed the linting error and the launch file typo, however i'm not sure how to test this as there is no logging when it receives input from teleop. Is there an activation through teleop i'm missing for cache mode or something? Initalises fine tho

tbh LGTM but someone more sciency should have a deeper look

FelicityMatt (approved):
LGTM

PR #1088, opened 2026-04-26
feat: Moisture Probe

### LINKED ISSUE #1017: Moisture probe
### **Context**

We are using the same moisture probe as last year, but we will be getting data over can this time instead of directly, so will have to add a sensor for this.

Should be similar to the BME sensor so see if you can share any code.

Existing code is called **hydraprobe** fyi, they've changed the term to moisture probe this year

### **TODO**

- create a pc2 system to publish this data to the gui.
  - the gui component should already exist but feel free to update as needed.

Tags for Notion Integration:
tag:software;tag:science

FelicityMatt:
## Description

Updated Hydraprobe / Moisture probe code

## Changelogs

- Removing move hydraprobe trigger code (no longer needed)
- Migrating code to PC2
- Removing some old services too

## Screenshots

image

## Memes

Tags for Notion Integration:
tag:science


V01DBREAKER (approved):
works on mine LGTM

PR #1090, opened 2026-04-26
feat: Cameras3 Upgrade 4

rotaryViper:
## Description

- Massive code refactoring
- Opengl accelerated pipeline
  - Denoising, edge detection and undistortion shaders
  - GPU accelerated image scaling, cropping and colour space changing
- rosimage integrated into pipelines
- Depreciating h264 and av1 software encoding

To do:
- Test on orin dev kit
- Modify resolution verification into a resolution warning system


src/ros/cameras/cameras3/default.nix
```diff
@@ -68,8 +75,26 @@ buildRosPackage {
     '';
 
   preFixup = ''
-    export GST_PLUGIN_PATH="${gst-bridge}/lib:$GST_PLUGIN_PATH"
-    wrapGApp "$out/lib/cameras/camera_streamer_service"\
-      --prefix GST_PLUGIN_PATH : "${gst-bridge}/lib"
+    set +e
+    # Detect GPU type (Intel or AMD)
+    GPU_TYPE=$(lspci | grep -iE "VGA|3D" | grep -i -E "Intel|AMD" | awk '{print $0}')
```
Redecorating, line 80: I don't think this is pure. Is it possible to set the env vars so the same build will let opengl try both mesa and nvidia? If hydra builds this it'll find an nvidia gpu but then the laptops with intel/amd will download that build and might not work.

Redecorating (changes requested):
aside from the gpu detection lgtm

src/ros/cameras/cameras3/default.nix
rotaryViper: It aint pure but there is no standards around gpus anyways, so this is a bodge that:
1. Assumes you are using a laptop and tries the default intel/ amd gpu driver with wayland mode
2. Assumes you are running headless and the library paths are already correct. And selects headless mode

PR #1092, opened 2026-04-27
Gui Dev Shell

V01DBREAKER:
## Description
This creates a nix-shell for development in the GUI
It does the following steps, builds all dependencies of GUI
copies node_modules into your local repo (no more running yarn install after gui-shell)
automatically links ros-types to your local repo (no more running gui-link)

creates new alias to run it `gui-dev-shell`

This does not modify original gui default.nix so gui should work as expected as it is untouched
This script also does not build gui-serve so it should be faster than running traditional gui shell

## Screenshots
image

## Steps to Test
`nova-shell -A pkgs.ros.nova-gui-dev-shell`
then try `gui-run`
gui should work as intended.

## Memes
image
This is why this PR exists, its so heavy yet we pull it twice ;-;

Tags for Notion Integration:
tag:nova-gui;tag:nixfiles

### DIFF
--- a/src/ros/nova-gui/nix/packages/gui/dev-shell.nix
+++ b/src/ros/nova-gui/nix/packages/gui/dev-shell.nix
@@ -1,0 +1,58 @@
+{ mkShell
+, config
+, nova-gui
+, pkgs
+}:
+
+# This shell is designed for use in development and it is assumed the developer will run the gui using yarn
+# If looking for a shell with the gui fully built, use the default.nix instead (e.g for gui-serve)
+
+let 
+
+gui = nova-gui.overrideAttrs(oldAttrs :{
+  # disable building of gui-serve to make it faster
+  postUnpack = "";
+  buildPhase = ''
+    runHook preBuild
+    runHook postBuild
+  '';
+  installPhase = ''
+    runHook preInstall
+    runHook postInstall
+  '';
+
+  # add path to node_modules to output to be used in dev shell
+  configurePhase = oldAttrs.configurePhase + ''
+    mkdir -p $out
+    echo $node_modules > $out/node_modules
+    echo $ROS_TS_DEFINITIONS > $out/ROS_TS_DEFINITIONS
+  '';
+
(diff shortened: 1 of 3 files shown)


V01DBREAKER:
Just check it builds before PRing it. It should tho

src/ros/nova-gui/nix/packages/gui/dev-shell.nix
```diff
@@ -0,0 +1,58 @@
...
+in
+
+mkShell {
+  preBuild = ''
+    echo "Entering GUI shell..."
+  '';
+  buildInputs = [ gui ];
+
+  shellHook = ''
+    if ! diff -rq $(cat ${gui}/node_modules) \${builtins.toString ../../../nova-gui/node_modules} ${diff_ignore} >/dev/null; then
```
Redecorating, line 45: do changes to node_modules/.vite trigger this

V01DBREAKER: No it ignores it because this is where temp stuff does when gui runs so it will always be different from the nix store

V01DBREAKER: This ensures it only triggers the copy on an actual package change

V01DBREAKER:
can confirm it builds

src/ros/nova-gui/nix/packages/gui/dev-shell.nix
```diff
@@ -0,0 +1,58 @@
...
+      echo "Copying nix-store gui modules $(cat ${gui}/node_modules) to ${builtins.toString ../../../nova-gui/node_modules}"
+      cp -rf $(cat ${gui}/node_modules) ${builtins.toString ../../../nova-gui}
+      chmod -R 777 ${builtins.toString ../../../nova-gui/node_modules}
```
Redecorating, line 48: im more of a 755 guy but it doesn't really matter

Redecorating (approved):
nice

PR #1094, opened 2026-04-27
can_sleuth: more realistic blcmds

Redecorating:
## Description

make the spin rate 100hz to match blcmds. also slow down the velocity because its way faster than the real arm and idk why.


Redecorating:
don't tell anyone i merged this without asking

PR #1095, opened 2026-04-28
Ducket URDF

V01DBREAKER:
image

Pull and run with `~/Builds/master/bin/ros2 launch arm_bringup urdf.launch.py local:=True arm:=True old_arm:=False use_mock_hardware:=True rviz:=True`
(Doesn't need a rebuild to test but you should check this builds)

Ducket:
image

Tags for Notion Integration:
tag:rviz;tag:ready_for_review


Redecorating (approved):
works for me. you do have a commit from another branch here btw so you may want to rebase.

V01DBREAKER:
oh good catch, will fix now

PR #1098, opened 2026-05-01
feat: yolo topic

tytia:
## Description

Publish yolo detections onto a ros topic for the BT to access

## Changelogs

- Try to use WebGPU by default
- Publishes new custom Detection2D msg
- Uses `baseStationIP` instead of hardcoding `10.0.0.12`
- Use bifrost instead of roslib directly

### DIFF
--- a/src/ros/nova-gui/nova-gui/src/components/auto/ObjectDetection/YoloProvider.tsx
+++ b/src/ros/nova-gui/nova-gui/src/components/auto/ObjectDetection/YoloProvider.tsx
@@ -77,6 +77,12 @@ export function YoloProvider({ children }: { children: React.ReactNode }) {
     // Don't try to publish if the websocket is offline
     if (!isRosConnected) return;
 
+    const activeVideo = videoRefs.current[0]?.current;
+    if (!activeVideo) return;
+
+    const vidWidth = activeVideo.videoWidth || 640;   // Fallback to 640
+    const vidHeight = activeVideo.videoHeight || 480; // Fallback to 480
+
     const activeDetections = detections.flat();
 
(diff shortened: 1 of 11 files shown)


src/ros/nova-gui/nova-gui/src/components/auto/ObjectDetection/YoloProvider.tsx
```diff
@@ -1,36 +1,117 @@
...
+
+  // Refs to hold the ROS connection and Topic so they persist across renders
+  const rosRef = useRef<ROSLIB.Ros | null>(null);
+  const topicRef = useRef<ROSLIB.Topic<any> | null>(null);
+
+  // Initialize ROS Connection on Mount
+  useEffect(() => {
+    const ros = new ROSLIB.Ros({
+      url: "ws://10.0.0.12:9090", 
+    });
```
tytia: Can this be changed to not be hard-coded and do it like how Cartographer or how any other ROS thing in the GUI does it and use the IP set on the GUI iykwim

FelicityMatt (changes requested):
do not use roslib directly, instead follow this pattern to publish:

https://github.com/MonashNovaRover/nova/blob/master/src/ros/nova-gui/nova-gui/src/components/drive/DriveJoyWidget/DriveJoyWidget.tsx

you will need to add the topic to each file in the nova-gui/src/ros/topics folder

V01DBREAKER (changes requested):
Please either rename the messages or use github.com/ros-perception/vision_msgs/blob/ros2/vision_msgs/msg
that way theres not duplicate msg names

PR #1110, opened 2026-05-05
feat: Update aliases to use ~/Builds/active

tytia:
## Description

Allows you to do `set_active <your_build_path>` and all aliases would be launched from your build

The `~/Builds/active` symlink resets to `~/Builds/master` on reboot

### DIFF
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -214,24 +214,25 @@ in
           # Launch rover or payloads
           rover-help = "more ${cfg.nixfileDir}/doc/rover-help.md";
           launch-teleop = "echo 'DEPRECATED - Please use launch-teleop-drive instead for drive teleop'";
-          launch-teleop-drive = "~/Builds/master/bin/ros2 launch teleop_drive_joy teleop.launch.py";
-          launch-drive = "~/Builds/master/bin/ros2 launch drive_bringup drive.launch.py";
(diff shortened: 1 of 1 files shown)


Redecorating:
what resets it to master on reboot? maybe you need systemd tempfile to do it? also would it be better to put this in /run

Redecorating:
although idk how you do stuff on boot with ubuntu+nix on pi5.

tytia:
> what resets it to master on reboot? maybe you need systemd tempfile to do it? also would it be better to put this in /run

the active build config is stored in `/run/user/$UID/nova/active_build` and sourced in every new shell session if it exists, otherwise the active build is always set to `~/Builds/master` before that

image

you can run `nova_sh_help` for some details on it

tytia:
> although idk how you do stuff on boot with ubuntu+nix on pi5.

this is in home manager so i'd assume it works the same?

nixfiles/modules/home/macros/default.nix
Redecorating: I don't think this should be run every time - if you try to run something as another session of bash is opening there's a race condition where you may follow the incorrect symlink before the other session rewrites it. Maybe make some file stored in /run that ensure this line only runs for the first bash session each boot and is otherwise skipped

Redecorating (approved):
what you have in this pr looks good. it's just one race condition i noticed when looking at how the code already on master sets the active build symlink.

nixfiles/modules/home/macros/default.nix
tytia: ooo true good catch, will fix

PR #1111, opened 2026-05-05
feat: add more monitoring tools

tytia:
## Description

Add `nethogs`, `glances` and `usbtop` and moved all monitoring tools into their own module

- `nethogs` is a per-process network bandwidth monitor
- `glances` is a kinda general system monitor like btop but with some network info as well
- `usbtop` monitors raw bandwidth usage over usb

## Steps to Test

```
btop

glances

nethogs

usbtop
```

### DIFF
--- a/nixfiles/modules/home/default.nix
+++ b/nixfiles/modules/home/default.nix
@@ -6,6 +6,7 @@
     ./branding
     ./desktop
     ./macros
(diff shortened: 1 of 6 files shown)


Redecorating (approved):
pls add wireshark and tcpdump too

PR #1115, opened 2026-05-13
Auto Typing Pokey PC2

jacksonmacc:
## Description

Rewrote the pokey.py script to python control 2. Also updated various things that had changed, such as the linear actuator can id. The pokey speed and time parameters should be adjusted for the linear actuator limits.

## Changelogs

* added PC2 to auto-typing package
* rewrote pokey.py to use PC2
* updated can details

## Screenshots

You can't see it in rviz, but the candump looks like it should.

## Steps to Test

```~/Builds/master/bin/ros2 launch arm_bringup path.control.launch.py use_mock_hardware:=true```
```launch-typing```
also needs GUI, rosbridge and cameras3 running, with the arm_end_periscope camera streaming.

## Memes

image


binudakal (approved):
lgtm!

PR #1116, opened 2026-05-13
feat: sequencer modes

binudakal:
## Description

Adds the option to localise once to get all keys and go to each consecutively (in addition to the existing approach of relocalising after going to each key), and makes it selectable from the gui.

## Changelogs

- Replaced `move_to_start` param with `relocalise`
- Removed 'relocalise' param and made the setting sent through the request from the gui

## Steps to Test

1. start gui (will need to do gui-link), rosbridge, cameras3 with arm_end_periscope streaming
2. `~/Builds/master/bin/ros2 launch arm_bringup path.control.launch.py use_mock_hardware:=true` 
(or just `launch-path-control use_mock_hardware:=true` if you've rebuilt)
3. `launch-typing`


jacksonmacc (approved):
looks good!

PR #1117, opened 2026-05-15
feat: positional cache servos

### LINKED ISSUE #1112: Caches - Positional
### **Context**

Science now wants to use positional servos for their caches - same as last year.

### **TODO**

- Revive the old gui widget/s
- Create a PC2 system for the position caches - try and reuse other stuff.

Tags for Notion Integration:
tag:science

binudakal:
## Description

Updates the caches to use positional servos instead of continuous ones, and restores their preset functionality in the GUI along with twitching support. Also updates the `launch-science-urc` alias. CAN ids yet to be updated

## Screenshots

image

## Steps to Test

1. run gui and rosbridge
2. `~/Builds/master/bin/ros2 launch science_bringup urc.launch.py` OR `launch-science-urc` if you've rebuilt


src/ros/rover/science/science/science/urc/cache.py
```diff
@@ -1,95 +1,54 @@
...
-from python_control2 import PythonControl, Controller, Contexts, InterfaceCollection, Interface, HardwareInterface
-
-from python_control2.hardware_interfaces import CMDHardware, ContinousServoHardware
+from python_control2 import PythonControl
+from python_control2.controllers import PresetTwitchController
+from python_control2.hardware_interfaces import PositionalServoHardware
 from teleop_python_utils import Inputs
 
 
-class CacheController(Controller):
```
FelicityMatt, line 26: Honestly I would leave this file be and add a new file called like caches_pos or something just so we have this redundant code if science decides to switch back

src/ros/nova-gui/nova-gui/src/components/science/CacheControlWidget/CacheControlWidget.tsx
```diff
@@ -1,43 +1,85 @@
...
+interface CacheConfig {
+  label: string
+  positionService: RosService
+  twitchService: RosService
+  storeKey: string
 }
 
-export default CacheControlWidget
\ No newline at end of file
+const CachePanel: React.FC<CacheConfig> = ({ positionService, twitchService, storeKey }) => {
```
FelicityMatt: I am not a huge fan of the left/right caches in the same widgets I think we will be moving them a lot, and it will be very annoying to have to switch between them, so I liked the way we did it before where we had one widget for each cache, I think that this is def doable with the current widget, remove the left/right tab, make the twitch buttons icon only and then put two of the widgets side by side

FelicityMatt (changes requested):
some comments

FelicityMatt (approved):
lgtm

PR #1118, opened 2026-05-15
Fix: autotyping bringup

jacksonmacc:
## Description

(This is Kal's work)
Fixes the launch-arm script to use autotyping, with a button to switch into the autotyping mode on the joysticks.

## Changelogs

* adds a new empty control mode to the arm teleop modular configs, that switches to the path planner
* adds the path planner to the main launch script

## Steps to Test

run:
* ```launch-arm use_mock_hardware:=true```
* ```launch-teleop-arm```
* ```launch-typing```
* plus run gui, rosbridge and cameras3 with arm_end_periscope


BaileyChessum:
you'll need to update the nixfiles to include the correct commit hash for the new empty control mode in teleop modular for this to work

jacksonmacc:
yeah, we were using the local mode to test, but we were having trouble getting it to build. not sure if we were doing something wrong but i left a note on the teleop_modular PR with more info, it looked like just a one line fix.

binudakal (approved):
lgtm

PR #1121, opened 2026-05-17
feat: drive improvements

Glass2721:
## Description

Many improvements to drive given provided feedback; this should cover most, if not all, issues with drive.

## Changelogs
- add autolock widget to all cameras pages (special cased for science urc to change its default location)
- increase cmd vel command timeout to 1 second (too short previously)
- deslopify drive autolock notifications
- make blcmd status monitor show gate driver faults (without discriminating between different error numbers; all are just considered "gate driver fault") and have drive autolock recognise these can messages as errors (enabling it to automatically trigger locking when there are too many)
- fast pivot blcmd reset (via aliases that make service calls to blcmd status monitor)
- improve pivot blcmd reset by delaying when zero position is set (~3 seconds after reset; hopefully should work consistently and reliably now)
- slow down update rate of drive joint state broadcaster so gui running via gui-shell has enough time to process each message

## Memes

With reference to the firmware required to support automatic pivot blcmd resetting:
Screenshot From 2026-05-18 01-32-17
there are now 1 million and 1 blcmd firmware changes to be made on the plane.

Tags for Notion Integration:
tag:ready_for_review

### DIFF
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -339,6 +339,16 @@ in
           # use this as `can_viewer can0` for example to get "-c can0"
           can-viewer = "can_viewer";
           can_viewer = "~/Builds/active/bin/can_viewer -i socketcan -c";
+
+          # Drive
+          reset-flw = ''~/Builds/active/bin/ros2 service call /blcmds/blcmd_reset blcmd_interfaces/srv/BLCMDReset "{type: 1, id: 1}"'';
+          reset-blw = ''~/Builds/active/bin/ros2 service call /blcmds/blcmd_reset blcmd_interfaces/srv/BLCMDReset "{type: 1, id: 2}"'';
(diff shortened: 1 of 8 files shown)


TJ-Lee-M:
Please add steps to test 
Steps you've already done helps us know exactly what "thouroughly tested" means - Important for quickly identifying any test cases missed and helping ppl efficiently verify they can at least replicate your results

Any list of expected behaviour to check on rover when we get to the US also useful - This is a big enough scope that we will want it ready to test in the US ASAP rather than risk merging now having only tested in sim.

Looking good from a v cursory look over code - havent gotten to test yet :)

tytia (approved):
Looks pretty good to me atm, but I didn't really look at the GUI changes (cause idk GUI) and I'm not too familiar with BLCMD firmware, so I'm not sure I understood all the changes made to `blcmd_status_monitor.py`. However auto blcmd resetting is a feature that can be easily disabled if there are issues so I see no harm in merging it :)

With these changes, is the intention to make auto blcmd resets on by default for both manual and autonomous operation? Also could you explain a bit more on how delaying when the zero position is set and checking the pivot zero every 30s helps with reliability?

TJ-Lee-M (approved):

PR #1124, opened 2026-05-18
search zone overlay on cartographer gui

TJ-Lee-M:
## Description
search zone overlay on cartographer gui :+1: 
boolean resets on refresh, I think thats fine

## Changelogs
add showSearchZones & toggleShowSearchZones state
add geojson circle helper
add button to toggle show/hide for all points 

## Steps to Test
gui-rosbridge, gui-run, tileserver

## Memes
image


src/ros/nova-gui/nova-gui/src/components/maps/Cartographer/components/BottomOverlay.tsx
FelicityMatt: can we make this text shorter, either convert to an icon or just omit the "show", we are already quite tight on space in the bottom bar. can add a tooltip around if needed.

would be cool to just be icons with like a search icon even? if you do that you should also move it to the right side where the other icons are in the bottom bar.

FelicityMatt (commented):
one comment otherwise lgtm

src/ros/nova-gui/nova-gui/src/components/maps/Cartographer/components/BottomOverlay.tsx
TJ-Lee-M: image

PR #1128, opened 2026-05-21
fix: es camera preset names

Glass2721:
## Description

Removes duplicate preset name from ES camera configs, and renames them to make more sense, and be selectable from the GUI. Does not change camera configurations.


FelicityMatt (approved):
lgtm

PR #1136, opened 2026-05-24
feat: cameras3 blacklist + vis-spec pipeline

V01DBREAKER:
### Changes
Adds a blacklist option to the streamer node in Cameras3
Also configures a ros topic pipeline for the uv vis spec

uv vis spec now uses the ros topic instead of cv2.videoCapture function

### To do (In future):
Things to fix (less important and can be done after comp)
- Blacklisted cameras will still show up in the gui camera control panel
  - This is because the gui control panel uses the directory topic and not the streamer node's information so to fix this would be to create a service on the streamer node which the gui polls to see what cameras actually have an active webrtc endpoint
- I deleted the start and end camera stream buttons on the uv vis spec, we should add a button which opens a camera control panel for this camera
- the ros topic of a pipeline should be advertised on the gui camera control panel as well as printed in console. currently the only way to know is either read the config or ros2 topic list and guess
- The logging on urc_uv_vis_spec.py is not accurate to whats happening and there's nothing that checks for the camera etc so if it has the topic but nothing on it, it will spam the console

### To test:
- Run GUI 
  - `gui-dev-shell`; `gui-run`
  - `gui-rosbridge`
- Run the uv vis spec node 
  - `ros2 run science urc_uv_vis_spec.py`
- Open the GUI and watch as the cameras work simultaneously with the UV Vis Spec

### Meme:

Tags for Notion Integration:
tag:science;tag:cameras


V01DBREAKER:
btw seems to work even if GUI isnt actively receiving the camera stream itself which is nice

PR #1140, opened 2026-05-25
Radio and wireless links monitor TUI

Will1306:
## Description

Simple TUI to show the status of all radio devices and wireless links to the rover. It updates in realtime and uses pings to determine which links are up. Useful for debugging during setup or simply monitoring the status of the networks as the rover operates.

Run with alias: radio-status

CTO locked in.

### DIFF
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -211,6 +211,8 @@ in
           N2 = "ssh -C -Y nova@10.0.2.12";
           N3 = "ssh -C -Y nova@10.0.2.13";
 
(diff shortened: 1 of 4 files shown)


Redecorating (approved):
looks fine. just be aware that if you connect both radios to the switch in the rover you'd be able to ping 900mhz over 2.4ghz and vice versa so it'd appear as if both are up even if one is down. (but with kevin that's not the case)

TJ-Lee-M:
Yep all good, we'll be running with kevin (but if something unexpected causes us to use last years set up instead we'll keep that in mind)
Thanks for the reminder 🫡

TJ-Lee-M (approved):

PR #1147, opened 2026-05-30
feat: arm telemetry (for delivery)

Glass2721:
## Description

Shows arm telemetry (velocity and effort) for each joint* on the arm in the delivery cameras page.

*all joints, except for j5 and j6 are shown. Instead, they are replaced by "roll" and "pitch", which show the "effort" and "velocity" exerted by those joints on the diff wrist if they existed instead of j5 and j6.


Redecorating (approved):
looks fine to me

TJ-Lee-M (approved):
Tested over radios

PR #1156, opened 2026-07-11
Feat/faster git metadata

Redecorating:
## Description

TODO: time the improvement in evaluation time


V01DBREAKER (commented):
if its faster then i will approve

Redecorating:
theoretically it is but i haven't had a chance to measure the final improvement

PR #1159, opened 2026-07-13
feat: Nova CLI

FelicityMatt:
## Description

Something that has been cooking in my brain for a while now.

The intent behind it is to simplify some of the core commands that we run, while also making it accessible (e.g. easy to run for non-software members) and easily extendable and not super hard coded (like aliases).

I want this to be something that people use often and feel comfortable adding to whenever there is an action that would be good to simplify.

want to run ANY launch file -> nova launch
want to run ANY ros2 node -> nova run
want to change an env var or the active build -> nova env
want to build the current workspace and have it go to ~/Builds -> nova build
want to automatically checkout, pull and rebuild master -> nova rebuild-master
want to start a terminal launch script -> nova start
can't remember how to run can_sleuth -> nova sleuth
can't remember any of the above -> nova

oh and did I mention it has auto complete?

> [!NOTE]
> This PR has been written mostly with Claude, but the architecture and implementation has been heavily reviewed and fine tuned by myself :)
> - The test interface for run and launch was optimised by me to get clear test cases, everything else is generated and idk.
> - I like the current structure of the project (easy to add new commands) but feel free to rip it apart in future if you think of anything better.

## Changelogs

- Add a new CLI
- Update aliases to point to the CLI

## Screenshots

image

## Steps to Test

Pull the branch, nixos-rebuild switch, open a new terminal tab and have a play around!!

## Memes

image (1)


V01DBREAKER:
i love the build and the rebuilding master for a future pr can we replace all the launch scripts with eith a json or yaml file that the nova cli can run?

V01DBREAKER (approved):

nixfiles/packages/python/default.nix
```diff
@@ -14,4 +14,5 @@
   wmm-calculator = callPackage ./wmm-calculator { };
   pypcd4 = callPackage ./pypcd4 { };
   ros2-unbag = callPackage ./ros2-unbag { };
+  nova-cli = callPackage ../../src/other/nova_cli { };
```
Redecorating, line 17: I think this should be in src/other/nova_cli/default.nix same as how it is for can sleuth? I think this folder is more for packaging stuff outside of the nova repo. That being said im pretty sure the reolink breaks this rule ;-;

V01DBREAKER: I think this is fine as it is a python package

V01DBREAKER: That said i think we need an audit of the nixpackages

Redecorating:
you may want to change the rebuild master to use the bash alias through `bash -ic ws-build` or copy `${nova-build} -A pkgs.ros.nova-workspace --argstr git-metadata \"$(${../../../scripts/generate-git-metadata.sh})\""`

PR #1146, opened 2026-05-28
coordinate conversion script for cartographer

jaredlandau:
## Description

- script to convert between degrees/minutes/seconds and decimal degrees
- defaults to DMS --> DD as cartographer GUI only accepts DD, but has input and output flags for different formats
- can parse either positive/negative coordinates (e.g. "-XX YY ZZ") and also N/S/E/W coordinates (e.g. "XX YY ZZ S")
- can handle multiple coordinates at once (e.g. convert_coordinates.py "XX YY ZZ" "AA BB CC"), but will assume this to be a combined latitude + longitude string
- example usage below

## Changelogs
added src/other/convert_coordinates.py
added 'cc' alias for convert_coordinates.py in default.nix

## Screenshots
image
image

## Steps to Test
~/src/other/convert_coordinates.py "XX YY ZZ"
cc "XX YY ZZ"

## Memes
image

Tags for Notion Integration:
tag:autonomous;tag:fieldtest;tag:gps;tag:ready_for_review

PR #1160, opened 2026-07-13
coordinate conversion script for cartographer

V01DBREAKER:
## Description

- script to convert between degrees/minutes/seconds and decimal degrees
- defaults to DMS --> DD as cartographer GUI only accepts DD, but has input and output flags for different formats
- can parse either positive/negative coordinates (e.g. "-XX YY ZZ") and also N/S/E/W coordinates (e.g. "XX YY ZZ S")
- can handle multiple coordinates at once (e.g. convert_coordinates.py "XX YY ZZ" "AA BB CC"), but will assume this to be a combined latitude + longitude string
- example usage below

## Changelogs
added src/other/convert_coordinates.py
added 'cc' alias for convert_coordinates.py in default.nix

## Screenshots
image
image

## Steps to Test
~/src/other/convert_coordinates.py "XX YY ZZ"
cc "XX YY ZZ"

## Memes
image

Tags for Notion Integration:
tag:autonomous;tag:fieldtest;tag:gps;tag:ready_for_review

### DIFF
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -314,6 +314,7 @@ in
           launch-gps-base = "~/Builds/active/bin/ros2 launch nova_bringup gps_base.launch.py";
           launch-magnetometer = "~/Builds/active/bin/ros2 run electronics magnetometer.py";
           mast = "ssh -C nova@10.0.0.150";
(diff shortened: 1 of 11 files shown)


V01DBREAKER:
See #1146

PR #1163, opened 2026-07-13
Feat/faster git metadata

Redecorating:
## Description

make sure we don't copy the whole ~/nova folder to the nix store when we build (including .git and gitignored stuff like node modules)

make the hash produced by nova git metadata consistent even if your gitignored stuff and .git folder contents are different


Redecorating:
will need a small change to nova cli when that is done

Redecorating:
I have checked that the workspace evaluates to the exact same hash with and without 7e1cba5

V01DBREAKER (approved):
If it's faster merge it!

PR #1166, opened 2026-07-15
feat: End to End tests

### LINKED ISSUE #1175: Create unit test framework
### **TODO**

-

V01DBREAKER:
Its less unit test and more end to end test lol oops
## Description
A test framework based in Pytest that allows the following things:
- Test any python code
- Run the test as a part of nix build (can block the build if it doesn't meet your test requirements [a good thing]!)
- Test CAN output of your nodes
- Send and receive ros2 topics and services

To achieve this there is a package with a Mock version of JCAN which allows you to inspect the CAN bus without using vcan
wsl lovers rejoice! (Just me lol)

There is currently only one test but its a great example of what you can do, you can find it at:
`src/ros/rover/science/science/science/pytest/test_kiln_temp.py`

To run it, simply run ws-build!
To run the test manually enter a shell with nova-science `nova-shell -A pkgs.ros.nova-science` then run `checkPhase` then you can run pytest `pytest src/ros/rover/science/science/science/pytest`.

To add tests to science simply add them to the same folder!
To add tests to other packages have a read of `src/other/nova_testing/nova_pytest_framework/README.md`

## Questions for the reviewer
- Is this the best place to put these packages?
- Is this the best place to put the pytest folder in science? I want it separated from the tests folder in science
- Is there anything that needs explaining in the docs to make this easier to use
- What would you like to see in a user guide (that i might be writing after this)?

## Screenshots

image
Running manually

image
Running with ws-build

## Memes

# The V Model is just Waterfall followed by Inverted Waterfall
image

# also the fact that we don't have a tests label is crazy
image
The closest would be CI/CD or field test LOL

Tags for Notion Integration:
tag:ci/cd;tag:simulation


V01DBREAKER:
Can someone smarter than me figure out the collision i've created:
image
@Redecorating

nixfiles/tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+ - Package level tests:
+   - https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests
+   - These tests can be used to check basic integration of a package
+ - CheckPhase tests:
+   - https://nixos.org/manual/nixpkgs/stable/#ssec-check-phase
+   - These tests can be used to run unit tests that are included with the package
+
+
+# What is in this directory?
+There are currently two tests implemented in this directory both of which have been untouched since 2024...
```
Redecorating: The networking one I made this year and I also fixed the cameras one this year!

nixfiles/tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+Both tests must be run manually.
+
+### `./cameras-webrtc`
+This tested cameras2 with a demo GUI that used to be included with gstreamer-rs plugin to test webrtc
+Unknown if this will work with cameras3. This test should be updated to work with cameras3
```
Redecorating, line 22: yeah it'll need to be updated at some point to start cameras3 instead of the soon to be deleted cameras2 (but don't let that be a blocker to deleting cameras2)

nixfiles/tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+### `./networking`
+A simple networking test to check the networking setup from 2024.
+The setup is not what we have now and this test needs to be updated.
```
Redecorating: it tests the PRP setup but it does need updating a bit since the setup got simplified to not have vlans (i did this on the can vm test branch but that's not merged)

src/ros/rover/science/science/default.nix
```diff
@@ -28,23 +29,44 @@ buildRosPackage {
...
     python3Packages.minimalmodbus
-  ] ++
-  [
     nova-python-control
     nova-python-control2
     nova-input-interfaces
     nova-camera-msgs
     teleop-modular-python-utils
     nova-science-interfaces
+    nova-pytest-framework
```
Redecorating, line 49: i think this depends on mock-jcan so you get both jcan and mockjcan maybe

src/other/nova_testing/nova_pytest_framework/default.nix
```diff
@@ -0,0 +1,20 @@
...
+  name = "nova-pytest-framework";
+
+  src = builtins.path rec {
+    name = "nova-pytest-framework-source";
+    path = ./.;
+    filter = lib.novaSourceFilter [ ] path;
+  };
+
+  propagatedBuildInputs = [
+    mock-jcan
```
Redecorating, line 18: as nova pytest framework is a propagated input for science, and nova pytest framework propagates mock-jcan, science ends up having both jcan and mock-jcan in its python site packages.

Redecorating (commented):
not sure what the best way to work around the jcan conflict is

another aproach might be smth like
```
import mock-jcan
import sys
sys.modules['jcan'] = mock-jcan
import jcan
assert jcan is mock-jcan
```
not sure if that needs to run before other code tries to import it. I think if u have something like import xxx.yyy you'd need to do this for every yyy.

Redecorating:
does vcan not work in wsl?

Redecorating:
> Is this the best place to put the pytest folder in science? I want it separated from the tests folder in science

could even convert those to pytest tests maybe? idk if people run the non pytest tests much

src/ros/rover/nova_generic/python_control2/default.nix
```diff
@@ -20,6 +21,7 @@ buildRosPackage {
     rclpy
     pythonPackages.jcan
     teleop-modular-python-utils
+    nova-interfaces
```
FelicityMatt, line 24: noooo i want to kill this

src/ros/rover/science/science/science/arc/kiln.py
```diff
@@ -22,23 +22,25 @@
...
 import rclpy
-from rclpy.node import Node
 from python_control2 import PythonControl
+from python_control2.controller_manager import ControllerManager
 from python_control2.controllers import ThermalController
-from python_control2.hardware_interfaces import QCMDHardware, GenericSensorHardware
+from python_control2.hardware_interfaces import GenericSensorHardware, QCMDHardware
+from rclpy.node import Node
 
+command_service = "/science/thermal_command"
```
FelicityMatt: i don't love this, i wanted to move away from having param defined at the top of files, also for science i would just assume that can1 is used no need to code it at the top of every file, it will always be the same default for all the nodes and would only ever change in the launch file/params files.

FelicityMatt: should at the very least be all caps for a const

src/ros/rover/science/science/science/pytest/test_kiln_temp.py
```diff
@@ -0,0 +1,77 @@
...
+from ..arc.kiln import can_bus, command_service, data_topic
+from ..arc.kiln import main as kiln_node
+
+# CAN bus name and heater CAN ID's as configured in arc/kiln.py's `with_jcan()` /
+# `with_hardware(...)` calls - kept in sync with that file.
+LEFT_HEATER_CAN_ID = 0x0C1
+RIGHT_HEATER_CAN_ID = 0x0D2
+
+
+def test_kiln_temp(logger, setup_tester, setup_sut):
```
FelicityMatt: I like this interface, altho the test itself could be improved lol

I reckon a good pattern for nodes like this would to have a setup step in each test that initialises the node and can bus and then only the actual test logic is in each test.

FelicityMatt (commented):
some initial thoughts, mostly focussing on the developer experience of writing the tests.

src/ros/rover/science/science/science/pytest/test_kiln_temp.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
```
FelicityMatt: also i agree with what orlando said i reckon put these in the test folder, you can remove whatever is there it'll be out of date now, and no one touches it

V01DBREAKER:
My thoughts on your reviews:
"i wanted to move away from having param defined at the top of files" oh phew, okay as long as you dont mind it being hardcoded in the tests
The test itself is an example of what could be done, i didnt really understand the kiln so i just made a very simple test that checked for the existance of responses
Ideally it should only need mock_jcan for the tests and checkPhase so there's probably a better way to do it
I tried the sys modules approach but it doesn't seem to work well i would like to stick to whats been done

V01DBREAKER:
Also the PR seems to have git metadata build errors? not sure why

Redecorating:
> Also the PR seems to have git metadata build errors? not sure why

is that on hydra? I think i know why -- fixed on master now

Redecorating:
```
diff --git a/src/other/nova_testing/nova_pytest_framework/default.nix b/src/other/nova_testing/nova_pytest_framework/default.nix
index 8ddac8773..9442b988c 100644
--- a/src/other/nova_testing/nova_pytest_framework/default.nix
+++ b/src/other/nova_testing/nova_pytest_framework/default.nix
@@ -14,7 +14,7 @@ buildPythonPackage {
   };

   propagatedBuildInputs = [
-    mock-jcan
+    #mock-jcan
     pytest
   ];
 }
```
this is enough to get rid of the collision error. 

I think the collision happens to be ok because nix decides to keep the real jcan instead of the mock but idk if that could randomly change in future so I think we should make sure it doesn't collide.

src/ros/rover/nova_generic/python_control2/default.nix
V01DBREAKER: will leave it in for now as pythoncontrol2 does have an import for this package which results in errors.

src/ros/rover/science/science/science/arc/kiln.py
V01DBREAKER: addressed, moving away from param at top of file

src/ros/rover/science/science/science/pytest/test_kiln_temp.py
V01DBREAKER: added a common fixture for kiln as an example + more docs

V01DBREAKER:
fixed the collision, addressed felicity's concerns

nixfiles/tests/README.md
V01DBREAKER: i fixed the readme issues caused by my ignorance

src/ros/rover/science/science/science/pytest/test_kiln_temp.py
V01DBREAKER: moved them to archive

V01DBREAKER:
Hello please rereview, i addressed ur reviews @FelicityMatt @Redecorating

src/ros/rover/science/science/science/tests/test_kiln_temp.py
```diff
@@ -0,0 +1,91 @@
...
+    spin_node_on_timer,
+)
+from science_interfaces.msg import ThermalData
+from science_interfaces.srv import ThermalCommand
+
+from ..arc.kiln import main as kiln_node
+
+# CAN bus name and heater CAN ID's as configured in arc/kiln.py's `with_jcan()` /
+# `with_hardware(...)` calls - kept in sync with that file.
+CAN_BUS = "can1"
```
FelicityMatt, line 19: i'd be extra lazy and hardcore these in place since they're only used once but it's fine

FelicityMatt (commented):
the science test looks good to me, i like the pattern and think it will do well, id get someone else to review the actual implementation though as well, as i did not review

Redecorating (approved):
had another read through, looks fine. Let's merge and people can start writing tests!

PR #1168, opened 2026-07-16
nixfiles: hydra: correct input name

Redecorating:
this (i think) fixes the hydra build status being integrated into github!

## Memes

the errors will continue until the vomiter improves
image
image

### DIFF
--- a/nixfiles/modules/nixos/common/ci/master/default.nix
+++ b/nixfiles/modules/nixos/common/ci/master/default.nix
@@ -105,6 +105,7 @@ in
           <githubstatus>
             jobs = nova:workspaces(?:-pr-.*-\d+)?:(?!.*-inputs).*
             useShortContext = 1
+            excludeBuildFromContext = 1
             inputs = nova-monorepo
             ${builtins.concatStringsSep "\n" (map
               (repo: "inputs = ${repo}")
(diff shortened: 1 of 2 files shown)


FelicityMatt (approved):
omg omg omg

are you seriously saying that i've tried to get this working for like a year and it was one stupid string i had to change

FelicityMatt:
but also is there a way to make the names of the checks a bit nicer?

Redecorating:
maybe excludeBuildFromContext but im not too sure what it does

https://github.com/NixOS/hydra/blob/c5e3a66c668a1c9858e1d8e9b22db7ffb9b9a6e0/subprojects/hydra/lib/Hydra/Plugin/GithubStatus.pm#L45

V01DBREAKER:
maybe we can consolidate it in some way so it just shows if the entire nova-workspace is built correctly or not?
Also is it possible to run this for docs and the other jobsets as well (if the code changes those jobsets only)

Redecorating:
> maybe we can consolidate it in some way so it just shows if the entire nova-workspace is built correctly or not?

not without patching hydra (not as bad as it sounds, it's one perl script you'd have to edit). github does hide the successful ones tho so it only shows the failed ones till you open the dropdown for the successful ones



> Also is it possible to run this for docs and the other jobsets as well (if the code changes those jobsets only)

yes but not only if it changed specific code

PR #1169, opened 2026-07-17
feat: DELETE python_control_old

FelicityMatt:

PR #1170, opened 2026-07-18
Feat/move stuff

Redecorating:
## Description

simplify our directory structure in src a bit.

`src/ros/nova-gui` -> `src/nova-gui`
`src/ros/rover/*` -> `src/ros/*`

why? Because I felt it was weird for cameras to be lonely. Gui is separate from ros because its mainly react.

this means we have

src/ros: ros stuff
src/other: other stuff
src/nova-gui: react gui stuff

also remove the nix code that dynamically discovered the default.nix's in src as we don't need that now that we have a monorepo and it's more clear how nix imports stuff this way.

## Why now orlando?

because we are at the start of the cycle, I wanted to do this ages ago but we won't get another chance till end of cycle.

If you really don't like it we can just take the first commit here (explicitly including all the stuff in src).

## My branch is messed up by this!!
```
git checkout your branch.
git merge this branch
git status # look for the red ones.
# deal with git's complaints
```

in the case of #1169 , I tested this and it did complain, but I just had to confirm that I wanted to delete the files that had been moved (so `git rm <all the files it asked about>; git commit` was all I had to do. Ask me if you need help.

## Memes

My most controversial PR yet (more to come...)


V01DBREAKER:
does this allow the gui to see ros still? i know packages in src/other dont get access to packages like pythonPackages in nix.

Redecorating:
it shouldn't impact the gui being able to see rosbridge and stuff. I updated the path in the gui-link alias.

image
fwiw can sleuth is in other and can access pythonPackages

V01DBREAKER:
readmes should also be updated to reflect new structure

PR #1181, opened 2026-07-22
📯 Unity sim 

vic-bart:
## Description

A long time coming, Unity is finally here to replace Gazebo. The hope is that the greater graphical quality and ubiquity of Unity over Gazebo will make it easier to develop a simulation for use by operators and the wider team, as well as in outreach events. 

## Changelogs

- Replaced Gazebo with Unity for simulating the auto stack.
- Added packages for nova-unity-sim (packaged release of unity-sim build), ros-tcp-endpoint (TCP bridge for Unity x ROS2 messages).
- Increased local_costmap.transform_tolderance from 0.2 -> 0.4 seconds (to prevent local_costmap not working).
- Refactored drive_bringup drive.launch.py to always use the can bus.
- Fixed TF tree for rover legs and diffbar.
- Increased the total number of nodes for Cyclone DDS from 32 (Jazzy default) to 100.
- Added C# VSCode extension.

## Screenshots

Screenshot From 2026-07-22 16-55-57
Screenshot From 2026-07-22 16-59-50

## Steps to Test

To test the stack, `launch-sim` or `ros2 launch auto_bringup sim.launch.py`.
To test with the Unity editor, install unity-sim and play the scene with the above arguments appended with `release:=false`.

## Memes

I have a whole bunch :grin: 
Screenshot From 2026-07-18 23-02-30
Screenshot From 2026-07-18 22-56-06
Screenshot From 2026-07-18 10-42-06
Screenshot From 2026-07-10 20-10-25
spinny


src/ros/rover/auto/auto_bringup/default.nix
```diff
@@ -106,7 +110,9 @@ buildRosPackage rec {
       demo-nodes-cpp
       pcl-ros
       direct-visual-lidar-calibration
-      nova-auto-start;
+      nova-auto-start
+      ros-tcp-endpoint
+      nova-unity-sim;
```
Redecorating: on hydra: `error: i686 Linux package set can only be used with the x86 family.`

you can do something similar to what they have here https://github.com/NixOS/nixpkgs/blob/fd1462031fdee08f65fd0b4c6b64e22239a77870/pkgs/os-specific/linux/displaylink/default.nix#L14 to make sure it doesn't include unity sim when building auto bringup for arm

vic-bart: Weird, I thought I had already done this.

PR #1186, opened 2026-08-01
nixfiles: remove default paths

Redecorating:
as we have a monorepo, we no longer need to dynamically discover where all the nova repos are

this makes it clearer how everything gets imported.


src/other/default.nix
```diff
@@ -0,0 +1,8 @@
+{
+  imports = [
+    ./can_sleuth
```
FelicityMatt, line 3: do you need the cli here too? or is that imported by another path (and if so should we change the import path?)

FelicityMatt (approved):
one comment but as long as that's accounted for lgtm

src/other/default.nix
Redecorating: oops! I will fix that now.

PR #1219, opened 2026-08-21
fix: nova cli main

V01DBREAKER:
Since we renamed master to main we need to update nova cli so that it can use main instead of master for the rebuild-master command. Also renamed any instance of master to main so all the tests and examples use main now.

Tested locally on mine and works fine.
To test either checkout and nixos rebuild or just run the nova command from your build folder
`nova build main`
`nova rebuild-main`
`nova` - check help message changed

Tags for Notion Integration:
tag:software

### DIFF
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -316,7 +316,7 @@ in
           launch-magnetometer = "~/Builds/active/bin/ros2 run electronics magnetometer.py";
           mast = "ssh -C nova@10.0.0.150";
 
-          # Master build binaries
+          # Active build binaries
           mros2 = "~/Builds/active/bin/ros2";
           mrviz2 = "~/Builds/active/bin/rviz2";
           mrviz = "~/Builds/active/bin/rviz2";
--- a/nixfiles/modules/home/macros/default.nix
+++ b/nixfiles/modules/home/macros/default.nix
@@ -99,7 +99,7 @@ in
         if [ -f "/run/user/$UID/nova/active_build" ]; then
           . "/run/user/$UID/nova/active_build"
         else
-          ln -sfn "$HOME/Builds/master" "$HOME/Builds/active"
+          ln -sfn "$HOME/Builds/main" "$HOME/Builds/active"
         fi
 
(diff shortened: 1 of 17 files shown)


src/other/nova_cli/nova_cli/commands/__init__.py
```diff
@@ -23,7 +23,7 @@
     'build': BuildCommand,
     'env': EnvCommand,
     'launch': LaunchCommand,
-    'rebuild-master': RebuildMasterCommand,
+    'rebuild-main': RebuildMasterCommand,
```
FelicityMatt: rebuild master command should main now

src/other/nova_cli/nova_cli/commands/rebuild_main.py
```diff
@@ -21,16 +21,16 @@
 
 
 class RebuildMasterCommand(Command):
```
FelicityMatt: ^

FelicityMatt (commented):
one missing replace otherwise lgtm

FelicityMatt (approved):
lgtm

nixfiles/modules/home/macros/default.nix
```diff
@@ -336,8 +336,8 @@ in
 
           # Science
           predict-shell = "nom-shell ~/nova/src/other/ilmenite_ml"; # please come up with a more descriptive and less generic alias
-          run-spec = "~/Builds/master/bin/ros2 run science urc_uv_vis_spec.py";
-          run-theta = "sudo -E LANG=C ~/Builds/master/bin/ros2 run science urc_theta_360_cam.py";
+          run-spec = "~/Builds/main/bin/ros2 run science urc_uv_vis_spec.py";
+          run-theta = "sudo -E LANG=C ~/Builds/main/bin/ros2 run science urc_theta_360_cam.py";
```
binudakal, line 340: Should these be active as well?

binudakal (approved):
lgtm just 1 comment

nixfiles/modules/home/macros/default.nix
V01DBREAKER: nah just leave for now cause they werent active before

PR #1225, opened 2026-08-27
Fix/x86 hydra

Redecorating:


rotaryViper:
We should probably add the nix community cachix, but apart from that lgtm

rotaryViper:
Reduce the priority of flox substituter, and make sure our substituter is the lowest priority of all

V01DBREAKER (approved):
Lgtm, I'm surprised you have to explicitly set the branch as the main branch should be default

Redecorating:
after merge the last box on this page can be changed to point to main https://hydra.novarover.space/project/nova/edit

Issue #1183, opened 2026-07-27
Automate updating Nix package revisions


vic-bart:
### **Context**

We use Nix/NIxOS to pin versions of package dependencies to prevent API changes during development. Previously, this would be manually updated every several months to get new upstream features, however the accumulated API changes create a great deal of work (see #631, ironically one year ago today). Automating this process using an LLM that can automatically create a new branch, update package dependencies, resolve new build errors from API changes, and open a PR, on a weekly or monthly basis would save time.

### **TODO**

- [x] Research
- [x] Implement
- [x] Test

Tags for Notion Integration:
tag:autonomous;tag:software;tag:nixfiles;tag:ci/cd;tag:qol

Redecorating:
btw there's a github actions workflow that's in the wrong spot (nixfiles/.github) that used to do this although last i heard it's broken

vic-bart:
# Goals
- [X] Set up opencode, see https://opencode.ai/docs/agents/.
- [X] Get opencode to ws-build and tell me about the error. Took >20K tokens. Correctly determined the cause of the error, and actually proved it as well (showed where in the code the patch being applied already existed, and therefore was redundant and needed to be removed). It parsed the entire output of ws-build; this will eventually get incredibly long, so I need to figure out how to only get it to ingest the error. Maybe a subagent to actually run the command, parse the error, and give the error back to the primary agent. Setting up a subagent to run the same command cut the token use of the primary parent agent to ~11K. Could probably go further using something like RTK, but it doesn't have any Nix support at the moment, see https://github.com/rtk-ai/rtk/pull/1046.
- [X] Same as above, but get it to fix the error itself.

# Notes
- My initial thoughts are like, write a python script with some LLM API that, then I run it on the Workshop PC, it will: checkout a new branch, get the latest nix-ros-overlay revision (and the recommended nixpkgs revision), ws-build, parse the output error, determine the fix, apply the fix, then ws-build again, and loop that until ws-build is successful, then submit a PR, at which point the python executable finishes. It would also be nice if it documented it's thinking process; like each successful change is committed with a detailed breakdown of what the issue was and how the change fixes it. 
- So I came across "claude code" once or twice, and it seems to be a Claude CLI API that does exactly what I want, nice. 
- https://www.reddit.com/r/claude/comments/1ryo0mo/please_explain_claude_code_and_its_potential_like/.
[...]

Issue #631, opened 2025-07-27
Updating nixpkgs revision


vic-bart:
**CONTEXT**
[nixpkgs](https://github.com/NixOS/nixpkgs) is the main repository of Nix packages for the Nix package manager. Our version of nixpkgs (the version used for Nova development environments) is [pinned](https://github.com/MonashNovaRover/nova/blob/d25456bc046fae00a5bb844f93ef8b763ac0a53d/nixfiles/revisions.json#L3) to a commit on August 19, 2024.

My onedrive package (OneDrive Client for Linux) is now out-of-date. This is unacceptable.
It is time to update our nixpkgs revision (at least to [this commit](https://github.com/NixOS/nixpkgs/commit/734a6421c46ff7c4864c6da0d5deb28fb2458ce3)). 

On a serious note, our nixpkgs are a year out-of-date and they should be updated for the upcoming design cycle, so I will be trying to push the versions as far as I can, bit by bit, until I hit an issue too frustrating to resolve.

Checkout `fix/nixpkgs-update` for progress.

Tags for Notion Integration:
tag:autonomous;tag:software;tag:nixfiles;tag:ci/cd

vic-bart:
Attempting to update nixpkgs to this [commit](https://github.com/NixOS/nixpkgs/tree/734a6421c46ff7c4864c6da0d5deb28fb2458ce3).

vic-bart:
<img width="1297" height="761" alt="Image" src="https://github.com/user-attachments/assets/c0255142-0f76-4697-9065-1ac87f3850c9" />

Made lots of progress 🙌 
Fixed a bunch of errors, mostly related to using deprecated functionality. 
Had a compatability issue with ultralytics; basically we're trying to build an old version, specifically v8.3.71, because ultralytics was only packaged on nixpkgs Mar 9, 2025. But this old version doesn't work with the newer version of numpy (v2.2.5), so it failed. But since we can now use the nixpkgs ultralytics (v8.3.143), I've decided to remove ours since there _should_ be no need for it.

vic-bart:
I've hit a wall with gz-msgs-vendor.
```
/nix/store/9ds850ifd4jwcccpp3v14818kk74ldf2-gcc-14.2.1.20250322/include/c++/14.2.1.20250322/bits/basic_string.h:3740:5: note:   template argument deduction/substitution failed:
[...]

PR #1228, opened 2026-09-02
Nix sisyphus

vic-bart:
## Description

See #1183.

We use Nix/NIxOS to pin versions of package dependencies to prevent API changes during development. Previously, this would be manually updated every several months to get new upstream features, however the accumulated API changes create a great deal of work (see #631). Automating this process using an LLM that can automatically create a new branch, update package dependencies, resolve new build errors from API changes, on a weekly or monthly basis, would save a considerable amount of time.

Introducing [OpenCode](https://opencode.ai/), an open-source framework for utilising LLMs from the CLI. This PR adds three things:
- OpenCode integration
- GitHub workflow to run OpenCode on a Github-hosted VM 
(not recommended; ws-build can take >20 GB of storage, while Github-hosted workflow runners are limited to 10 GB; using a self-hosted workflow runner would fix this, but it's strongly advised against on public repositories as it can be used to inject malicious code onto the local machine)
- Shell script and alias to run OpenCode on a local machine, with a [systemd automation](https://github.com/MonashNovaRover/nix-configs/blob/workshop-build-pc/sisyphus.nix).
(recommended; can be run using `nix-sisyphus` alias)

The workflow/script checks out to a new branch, updates nixpkgs and nix-ros-overlay to their latest respective versions (sourced from [nix-ros-overlay's flake.lock](https://github.com/lopsided98/nix-ros-overlay/blob/develop/flake.lock#L27), which is bumped every month), and then loops through ws-build -> fix errors -> commit fix -> ws-build until there are no more build errors remaining. 

There are still some problems with this approach. The script generates a flake.lock and removes nixfiles/secrets to prevent the LLM reading it, which should be reverted before merging. LLMs are also inherently probablistic; they can screw up, and just because ws-build runs, it doesn't mean it works (e.g. nova-unity-sim broke ws-build, so it decided to remove nova-unity-sim). Any work committed by the LLM needs to be analysed and tested before merging.

## Changelogs

- OpenCode integration
- GitHub workflow to run OpenCode on a Github-hosted VM 
(not recommended; ws-build can take >20 GB of storage, while Github-hosted workflow runners are limited to 10 GB; using a self-hosted workflow runner would fix this, but it's strongly advised against on public repositories as it can be used to inject malicious code onto the local machine)
- Shell script and alias to run OpenCode on a local machine, with a [systemd automation](https://github.com/MonashNovaRover/nix-configs/blob/workshop-build-pc/sisyphus.nix).
(recommended; can be run using `nix-sisyphus` alias)

## Steps to Test

- `nix-sisyphus`

## Memes

image

Tags for Notion Integration:
tag:autonomous;tag:software;tag:nixfiles;tag:ci/cd


V01DBREAKER (dismissed):
LGTM send it

PR #1229, opened 2026-09-02
feat: Updating Nix packages

vic-bart:
## Description

This PR was produced by Nix Sisyphus, and reviewed + modified + validated by @vic-bart.

It updates our nix-ros-overlay and nixpkgs to their latest revisions (sourced from nix-ros-overlay's flake.lock), and makes whatever modifications are necessary to evaluated nova-workspace.

## Changelogs

The following is a checklist of the changes proposed by Sisyphus that must be verified:
- [X] Updated Qt patch
The generated patch was correct; it was verified against a git diff of a modified nix-ros-overlay.
- [x] Removed librealsense2 patch
The patch was added to nix-ros-overlay on Sep 16, 2025 (see [commit](https://github.com/lopsided98/nix-ros-overlay/commit/5b0989871a8e6c95b34ea11e6a6cde45f2a469bf#diff-245d9147ccfe8926df2d4eef8cb91c53bf221cba633508ad75c5fd72db4d2888R222-R224)).
- [X] Updated nav2-behavior-tree patch
This vexes me. For context, I assume that this patch exists because it is necessary for nav2-mppi-controller. In fact, running `nix-store --query --tree` on nav2-mppi-controller shows nav2-behavior-tree as a direct dependency, but it's not added anywhere as a dependency in our nixfiles or on nix-ros-overlay, or even in the source code. So I'm not exactly sure why this is here. Git blame shows Josh added it back with the initial MPPI backport (see [here](https://github.com/MonashNovaRover/nova/commit/403c4f9731a997a1bafcc644f6642f31139de496#diff-4df551be183178ed2c3e249ec7de63f419c7ff8b00c3aa355db47766dba6441aR45-R62)).
[EDIT] nav2-mppi-controller is fine when the patch is removed, but nav2-core fails as its cmake can no longer target nav2-behavior-tree. Removing nav2-core backport patch fixes this; removing backport patches for nav2-core and nav2-behavior-tree lets nav2-mppi-controller evaluate. But does it work?
[EDIT] This is different to what I thought; it isn't that these patches have been merged upstream, they're from a different Nav2. For context, it seems Nav2 does not target development on the latest LTS version of ROS2, but just the latest version in general; these backports may not be necessary to have nav2-mppi-controller "functional", but it will be an older version lacking the improvements made to it since Kilted released in May 2025. 
TL;DR if MPPI works when tested, this patch should be sufficient for now, until someone is willing to either: A) identify which patches are needed to get the latest MPPI improvements backported to Jazzy, or B) package MPPI Lyrical for Jazzy.
[EDIT] MPPI seems to work. Difficult to test in Unity because the pointcloud output is very intermittent (this is a separate issue that needs fixing; lowkey might add Gazebo back in just for auto testing in the meantime since the pointcloud was more reliable). Did encounter this case where the rover got too close to an obstacle to back away and spammed:
```
[INFO] [controller_server]: Optimizer reset
[INFO] [controller_server]: Optimizer reset
[WARN] [controller_server]: Optimizer fail to compute path
```
image

- [X] Update nav2-mppi-controller patch
For context, refer to the dotpoint above. A large portion of this is the LLM generating and executing Python scripts during the patch phase to change the source code :skull:. It also gave up and rewrote the CMakeLists.txt rather than create a patch (I think the LLM hates patches). As mentioned before, we should change the way we use MPPI in this repo, preferably finding a way to compile the Lyrical version on Jazzy (which isn't a crazy concept, e.g. some packages like FAST-LIVO2 on Humble work out of the box on Jazzy). If that fails, this can be revisited. At the moment, verifying and reimplementing this patch isn't a priority so long as MPPI performs well enough when tested.
- [X] Updated fastrtps
fastrtps is overrided to build the same version it's already pinned to on nix-ros-overlay, so that can be removed. fastrtps builds and runs without the patch as well; Seems like sisyphus hallucinated a fix :eyes: 
- [X] Updated ntrip-client-node patch
[EDIT] This comment deleted for some reason, thanks GitHub. TL;DR patch wasn't needed, upstream builds just fine.
- [X] Updated json-tricks
[EDIT] This comment deleted for some reason, thanks GitHub. TL;DR patch wasn't needed, single failing test can just be disabled.
- [X] Updated mavproxy
Despite defining the version to be 1.8.74, MAVProxy's CLI provides 1.8.71, so dontCheckVersion = true; is the right call.
- [X] Updated jcan (in two places? [EDIT] Looks like jcan is packaged twice, probably for Python and C++)
Seems like Sisyphus was telling the truth, see [here](https://gcc.gnu.org/gcc-15/porting_to.html#header-dep-changes), as of GCC 15+, <cstdint> is required for certain types.
- [X] Removed nova-unity-sim
Fixed this by replacing steam-run-free with nix-ld and providing the runtime dependencies to the LD_LIBRARY_PATH. Since this requires setting `programs.nix-ld.enable = true;` on everyone's devices, might re-implement this to use autoPatchelfHook instead of nix-ld. [EDIT] Reimplemented to use autoPatchelfHook.
- [X] Updated reolink
LLM is right again, see [here](https://github.com/NixOS/nixpkgs/pull/513497), mkYarnPackage was removed on April 26 and cannot be used. The changes closely follow the example template, see [here](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/javascript.section.md#yarn-v1-javascript-yarn-v1), and skipping the build phase is suitable as the source files don't require it.
- [X] Updated ros-typescript-generator
LLM is kind of right. mkYarnPackage was removed and has been replaced with yarnConfigHook, yarnBuildHook, and yarnInstallHook (see [here](https://github.com/NixOS/nixpkgs/blob/e72e4f299401a3689d4b3d5fc6496b11db7064eb/doc/languages-frameworks/javascript.section.md#yarn-javascript-yarn)). yarnConfigHook contains `export HOME="$(mktemp -d)"`, and yarnBuildHook contains an equivalent to yarn build --offline, so there's no need for any overrides.
Note that while v1.7.0 builds fine, I have bumped it to the latest v1.10.0 release. This could also be added to nix-ros-overlay as a PR.
- [X] Updated lap
pyproject = true
- [X] Updated minimalmodbus
pyproject = true
- [X] Updated mkdocs-safe-text-plugin
This is strange. pytestrunner was changed to pytest-runner Oct 29, 2025, but the LLM decided to rename pytest-runner back to pytestrunner? This isn't even a dependency of nova-workspace. [EDIT] Lol, changing it to pytest-runner hits you with the `Function called without required argument "pytest-runner" at nixfiles/packages/python/mkdocs-safe-text-plugin/default.nix:7, did you mean "pytestrunner"?`. [EDIT] So pytestrunner was an alias for pytest-runner, which was removed back in 2024? [EDIT] This package does NOT even build on main, so it was always broken :disappointed: LLM was wrong, the correct fix is to replace pytest-runner with pytestCheckHook (see [here](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#using-pytestcheckhook-using-pytestcheckhook)).
I also bumped it from 1.5.1 to 1.6.1, because 1.5.1 required pytest-runner + newer = better.
- [X] Updated ros2-unbag
LLM is kind of wrong-ish again. v1.2.3 is out-of-date, so it's requiring some older versions of packages. Now, if we were using flakes, I could just define an older nixpkgs as the input to this file, and we're done. But we don't use flakes. So instead, I'm going to bump it to v1.3.1 (latest) and repackage it (build system transitioned from wheel to cmake). [EDIT] Had to move derivation to packages/ros since it requires ament-cmake from packages/ros inherited callPackage, which isn't passed to packages/python inherited callPackage. [EDIT] This was added to nix-ros-overlay back on Aug 22, so our local derivation can be removed. 
- [X] Updated data-gradients
Why did Sisyphus rename this to imagededup? [EDIT] This was pretty messy; redundant patching, no build system configured, etc., required some clean up.
- [X] Updated super-gradients
Seems like the LLM hit on the tried and true strategy of, if dependencies aren't being satisfied, ignore the dependencies :sob: Bumping this from 8.0.146 (it's actually 3.4.0?) to 3.7.1, but that hasn't been updated for 2 years. Maybe Sisyphus was right, seems like this will require significant upstream work to be usable again. If it doesn't work with the newer dependencies, consider converting this to a flake to use the relevant versions (or you can manually define each one in the file).
- [X] Updated depthai
Looks good, just bumped the versions and hashes to what's available on the mirror. Replaced hostPlatform to remove evaluation warning.
- [X] Updated livox-sdk2
Looks good; it's CMake requires a minimum of v3.0, whereas modern CMake prohibits anything below v3.5, so it forces it to build with v3.5, which doesn't seem to error. Realised this should also pin a specific version release, and not an arbitrary revision; bumped to v1.3.1.
- [X] Updated spatio-temporal-voxel-layer
Fix was good, but we're building a version almost identical to what nix-ros-overlay has, so removing ours.
- [X] Updated vikit-common
Fix was reasonable, but instead of adding -DCMAKE_POLICY_VERSION_MINIMUM=3.5, it patched the CMake to require v3.5; made it use the flag instead to avoid modifying the patch.
- [X] Updated yolo-ros
Good fix. Decided to bump this from 4.0.1 to 4.7.1.
- [X] Updated can-sleuth
Explicitly added setuptools.
- [X] Updated nova_cli
Explicitly added setuptools.
- [X] Updated coms_utils
Explicitly added setuptools and renamed can -> python-can (happened Aug 13, 2024, see [PR](https://github.com/NixOS/nixpkgs/pull/334041))
- [X] Updated cameras3
Removed gst-vaapi, as it no longer builds upstream (see [commit](https://github.com/NixOS/nixpkgs/commit/b812bc2a4a1597b73bc9a5422318fe6bcc9bde34)).
- [X] Updated gui
See the above concerning the deprecation of mkYarnPackage. Cleaned up the implementation to change as little as possible. The deps folder is no longer present; I've done my best to replace it, but someone from software who knows how to build a Yarn package should probably look it over.
- [X] Added rosTypes.ts?
This gets auto-generated when nova-gui is evaluated and symbolically linked when gui-link is entered; can be removed.
- [X] Updated gimbal_cam
pyhtonPackages renamed to python3Packages.
- [X] Updated nova_detection_overlay
pyhtonPackages renamed to python3Packages.
- [X] Updated nova_object_localisation
pytorch renamed to torch.
- [X] Updated auto_start
pyhtonPackages renamed to python3Packages.
- [X] Updated lattice_primitive_generator
pyhtonPackages renamed to python3Packages.
- [X] Updated nova_utils
pyhtonPackages renamed to python3Packages.
- [X] Updated dgnss
pyhtonPackages renamed to python3Packages.
- [X] Updated electronics
pyhtonPackages renamed to python3Packages.
- [X] Updated excavation_construction
pyhtonPackages renamed to python3Packages.
- [X] Updated blcmd_utils
pyhtonPackages renamed to python3Packages.
- [X] Updated cmd_utils
pyhtonPackages renamed to python3Packages.
- [X] Updated hardware_interfaces
pyhtonPackages renamed to python3Packages.
- [X] Updated python_control
pyhtonPackages renamed to python3Packages.
- [X] Updated python_control2
pyhtonPackages renamed to python3Packages.
- [X] Updated science
pyhtonPackages renamed to python3Packages.

The following is a checklist of the changes Sisyphus missed:
- [X] Updated nav2-behavior-tree
GetCurrentPost and ArePosesNear were backported to Jazzy on Dec 6, 2025 (see [PR](https://github.com/ros-navigation/navigation2/pull/5755)), causing this error:
```
[FATAL] [bt_navigator]: Failed to create navigator id navigate_to_pose. Exception: ID [GetCurrentPose] already registered
[FATAL] [bt_navigator]: Failed to create navigator id navigate_to_pose. Exception: ID [ArePosesNear] already registered
```
Fixed by removing our local versions.

## Steps to Test

`ws-build`

## Memes

Tags for Notion Integration:
tag:autonomous;tag:software;tag:nixfiles;tag:ci/cd;tag:dependencies


V01DBREAKER (commented):
gui builds as expected so approved from me. Just remove the gui-dev-shell and move the default.nix from nix/packages/gui/default.nix up to the default.nix (all in src/ros/nova-gui)

V01DBREAKER:
actually maybe not the moving up part

V01DBREAKER (dismissed):
ah ill do the stuff i mentioned in a diff branch
LGTM if its building on ur end (cant seem to build it on mine due to limited memory)

nixfiles/overlay/ros/backports.nix
```diff
@@ -160,6 +160,295 @@ self: super:
...
+          patchPhase = ''
+            runHook prePatch
+            for patch in $patches; do
+              echo "Applying patch $patch"
+              patch -p2 --batch < "$patch" || echo "WARN: patch had partial failures, continuing..."
+            done
+            runHook postPatch
+          '';
+
+          postPatch = ''
```
Redecorating, line 175: thats crazy, big pickle

src/ros/cameras/cameras3/default.nix
```diff
@@ -48,7 +48,6 @@ buildRosPackage {
     gst_all_1.gst-plugins-bad
     gst_all_1.gst-plugins-ugly
     gst_all_1.gst-libav
-    gst_all_1.gst-vaapi
```
Redecorating, line 51: did this need replacing with gst-plugin-va?

src/ros/nova-gui/nix/packages/gui/default.nix
```diff
@@ -1,49 +1,64 @@
...
-, nova-arm-interfaces
-, nova-input-interfaces
-, nova-cmd-interfaces
-, nova-interfaces
-, nova-camera-msgs
-, nova-science-interfaces
+{ 
+  buildEnv, 
+  fetchYarnDeps, 
+  lib,
```
Redecorating, line 4: please complain to the robot about changing the locations of the commas for no reason.

src/ros/rover/auto/auto_bringup/default.nix
```diff
@@ -48,7 +48,7 @@
   direct-visual-lidar-calibration,
   nova-auto-start,
   ros-tcp-endpoint,
-  nova-unity-sim,
+  nova-unity-sim,
```
Redecorating, line 51: whitespace?

src/ros/rover/auto/auto_bringup/default.nix
```diff
@@ -66,7 +66,6 @@ buildRosPackage rec {
     launch-ros
   ];
 
-  isx86 = (lib.systems.elaborate builtins.currentSystem).isx86_64;
```
Redecorating, line 69: does this break building on arm?

Redecorating (changes requested):
in particular, does adding unity sim unconditionally break building on arm? and if the gst-plugin-va needs adding add that.

~~Some whitespace and formatting changes seem unnecessary, maybe we can add to the instructions to avoid that sort of thing? But that's not too important~~

src/ros/rover/auto/auto_bringup/default.nix
Redecorating: nvm it was human whitespace so i don't mind it

src/ros/nova-gui/nix/packages/gui/default.nix
Redecorating: nvm it was you so i don't mind it

vic-bart:
> Just remove the gui-dev-shell

@V01DBREAKER Sure but any reason in particular?

vic-bart:
> and if the gst-plugin-va needs adding add that.

gst-plugin-va is already included under gst-plugins-bad, verified with `gst-inspect-1.0 va` (see [here](https://wiki.archlinux.org/title/GStreamer)).

Also I'll try to avoid combining reformatting + code updates in the same PR, apologies.

src/ros/cameras/cameras3/default.nix
vic-bart: gst-plugin-va is already included under gst-plugins-bad, verified with gst-inspect-1.0 va (see [here](https://wiki.archlinux.org/title/GStreamer)).

src/ros/rover/auto/auto_bringup/default.nix
vic-bart: Perhaps? It's been removed now.

nixfiles/overlay/ros/backports.nix
vic-bart: Lowkey MPPI is a menace, I reckon it would be easier to get the Lyrical version of MPPI building on Jazzy than to backport Lyrical changes to Jazzy, otherwise we end up with incredibly lengthy and convoluted changes like this, that will have to be rewritten again and again as updates are made.

